PR #1, opened 2016-01-05
Show absolute representation of less significant digits

dsydalziel:
Will properly format negative numbers (i.e. -10.600$ instead of -10.-600$), see http://discoverygc.com/forums/showthread.php?tid=134618


zibbizor:
Good find. I talked with FriendlyFire about this edit as it's a core edit and we're also merging it with the SVN.

PR #2, opened 2016-01-05
Change AUTOBUY_CARTITEM iCount to signed int

dsydalziel:
Handle negative iCount values properly (i.e. due to bot capacity nerf), see: http://discoverygc.com/forums/showthread.php?tid=134618 (it should have displayed -53 instead of 4294967243). This change will make AUTOBUY_CARTITEM->iCount consistent with CARGO_INFO->iCount.

NOTE: I don't have access to a test environment at the moment (I'm away for the holidays) so I haven't been able to test this change yet. It looks to me like it should be good, but if you have any doubts at all then don't merge it yet.

### DIFF
--- a/trunk/plugins/autobuy/Main.cpp
+++ b/trunk/plugins/autobuy/Main.cpp
@@ -244,7 +244,7 @@ void LoadSettings()
 struct AUTOBUY_CARTITEM
 {
 	uint iArchID;
-	uint iCount;
+	int iCount;
 	wstring wscDescription;
 };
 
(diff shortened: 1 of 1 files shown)


zibbizor:
in Freelancer, iCount is an unsigned integer, hence why we don't recast in FLHook. It's worth a shot, I will have to test before I merge this..

dsydalziel:
Finally got around to testing it, looks good:

![autobuybb image]
(http://i.imgur.com/dKdKA7p.png)

zibbizor:
Ok, that works too. Sorry I didn't get time to get around it. Thanks for testing it! Merging it now.

dsydalziel:
No probs, sorry for submitting an untested PR in the first place.

PR #3, opened 2016-02-28
PluginUtilities cleanup

dsydalziel:
As discussed. Some notes [here](https://github.com/dsydalziel/FLHook/wiki/PluginUtilities-Consolidation-and-Cleanup)


dsydalziel:
The commit author field seems to have gotten messed up on some of these, will re-submit.

PR #7, opened 2016-04-28
Avoid crashing if we can't write logs

### LINKED ISSUE #6: Avoid crash in AddLog due to not being able to create files
You might not be able to create logs e.g. if not running as an administrator
When this happens we should probably just warn in console or something.
http://discoverygc.com/forums/showthread.php?tid=134709&page=5

dsyalex:
Fixes #6 in core


zibbizor:
Good stuff m8

PR #11, opened 2016-05-04
Event cargo count check bug fix

dsydalziel:
Quick fix for [this](http://discoverygc.com/forums/showthread.php?tid=138771&pid=1774034#pid1774034).


zibbizor:
I've put this fix in the main development branch, it will be published here soon along with all other improvements. thank you though

PR #12, opened 2016-05-16
Cope with commas in event tracker data

dsydalziel:


dsyalex:
Why was this closed?

dsydalziel:
Oh hey. It worked (I think) but it was a bit of a hasty hack. The ideal solution was to encode the data, which I started working on around that time. Unfortunately that work is gone now as my hard drive went bang last year (and with it some commits I hadn't pushed).

dsyalex:
Do you remember what the topics of those unpushed commits were?

dsydalziel:
Pretty sure that was the only one of any use. Any others were most likely grotesque experimental abominations that should never have been unleashed upon the world anyway!

PR #17, opened 2017-04-17
Add /jettisonall command to jettison all cargo

dsyalex:
Reuses UserCmd_Process code from another plugin
Reuses a bunch of code from playercntl to do the actual dropping.

In minecontrol because I felt that's where it's really useful. We can move
it elsewhere if wanted.

http://discoverygc.com/forums/showthread.php?tid=146352


zibbizor:
Seems good, I rather have it in Alley plugin though with the other "utility" things. We could also add a keybind for it.

dsyalex:
Update: I've been looking at how you create a keybind. It looks like our existing code for it is buried in dsace so I probably won't be able to actually test the result.

PR #18, opened 2017-04-17
playercntl: Allow further locking of ships

dsyalex:
* Lock HpCM* (Cloaks, Cloak Disruptors) as well as cargo
* Allow for an extra lock level that restricts selling anything.

http://discoverygc.com/forums/showthread.php?tid=146200

### DIFF
--- a/Plugins/Public/playercntl_plugin/Main.cpp
+++ b/Plugins/Public/playercntl_plugin/Main.cpp
@@ -618,23 +618,18 @@ namespace HkIServerImpl
 					{
 						string hp = string(item->szHardPoint.value);
 						boost::to_upper(hp);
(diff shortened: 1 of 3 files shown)


zibbizor:
This is good, perhaps we could also lock the HpSpecialEquipment slots as these are bound to receive use.

dsyalex:
There's another ini problem with that, talking to the dev chat.

dsyalex:
Update: I've been having trouble convincing people to look into the issue I found. I'll go ahead and update this later anyway, though it may not be 100% effective for locking HpSpecialEquipment in discovery for now

dsyalex:
Turns out it got fixed and I wasn't notified (or if I was, I didn't receive it) - DGCRepository commit c65513db7241f5a7a3a0fbdc8c2aa5e2fbb782ed

PR #24, opened 2018-03-02
Added HkIsOnDeathMenu.

LazDisco:
Added a way to detect whether the person is dead in space. Useful check for preventing people from using dialogue box commands and softlocking their game.


zibbizor:
This is a very conditional scenario. If this is intended for your peronal usage, yeah I guess, but this isn't really a hook so I don't believe it should be in the main hook project as it could be misleading for uninformed developers.

@dsyalex should have a look at this as well.

dsyalex:
"This is a very conditional scenario."  - not sure what this means?

Am trying to figure out how "it could be misleading for uninformed developers." Do you think there may be some case where it doen't work as described?

It probably belongs near HkIsInCharSelectMenu in HkFuncTools.cpp instead, but...

zibbizor:
@dsyalex There could be a case we are not aware of matching these criterias and unrelated to being "dead".

If you think this is good enough, feel free to merge it.

dsyalex:
I think this is good enough after simply moving it to HkFuncTools.cpp, if someone finds a case in which it breaks we can easily reconsider.

PR #27, opened 2018-03-11
Getrep Plugin

LazDisco:
New Command: /rep - This command will display the full list of npc reputations for the player. It will list them in order of reputation, rather than faction name.

### DIFF
--- a/FLHook.sln
+++ b/FLHook.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26228.12
+VisualStudioVersion = 15.0.27130.2036
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Template", "Plugins\Public\__PLUGIN_TEMPLATE\Template.vcxproj", "{E7008BA7-4F75-4AF7-B25C-3A9A3287A88F}"
 	ProjectSection(ProjectDependencies) = postProject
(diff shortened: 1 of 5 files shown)


FLHook.sln
```diff
@@ -68,25 +72,21 @@ Global
 		Plugins\PluginUtilities\PluginUtilities.vcxitems*{81a35b95-1ced-4f58-a24c-e2dc47d43dd0}*SharedItemsImports = 4
 		Plugins\PluginUtilities\PluginUtilities.vcxitems*{81d33b95-1ddd-4f58-a24c-e2ec4a143dd0}*SharedItemsImports = 4
 		Plugins\PluginUtilities\PluginUtilities.vcxitems*{81d35b95-1cdd-4f58-a24c-e2ec47d43dd0}*SharedItemsImports = 4
+		Plugins\PluginUtilities\PluginUtilities.vcxitems*{8cbf9ef0-324a-4699-8a26-c2999b99b225}*SharedItemsImports = 4
```
dsyalex: It looks like we're adding more things than should be in this pull request to this file?

Plugins/Public/getrep/Main.cpp
```diff
@@ -0,0 +1,233 @@
...
+	{
+		float fRep = 0.0f;
+		string factionName = iter->first;
+		uint factionExternal = iter->second;
+
+		wstring Faction = HkGetWStringFromIDS(factionExternal);
+		uint playerRep = HkGetRep(wscCharName, stows(factionName), fRep);
+
+		factionReps[wstos(Faction)] = fRep;
+		string FactionVector = wstos(Faction);
```
dsyalex: unused

Plugins/Public/getrep/Main.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+	return true;
+}
+
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////
+//Calls
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+/** Clean up when a client disconnects */
+void ClearClientInfo(uint iClientID)
```
dsyalex: We're not storing any per-client info here so this is unnecessary

Plugins/Public/getrep/Main.cpp
```diff
@@ -0,0 +1,233 @@
...
+	wstring wscCharName = (const wchar_t*)Players.GetActiveCharacterName(iClientID);
+
+	for (map<string, uint>::iterator iter = factions.begin(); iter != factions.end(); iter++)
+	{
+		float fRep = 0.0f;
+		string factionName = iter->first;
+		uint factionExternal = iter->second;
+
+		wstring Faction = HkGetWStringFromIDS(factionExternal);
+		uint playerRep = HkGetRep(wscCharName, stows(factionName), fRep);
```
dsyalex: this actually returns an HK_ERROR, maybe we should detect it erroring and avoid updating factionReps

Plugins/Public/getrep/Main.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include <algorithm>
+#include <FLHook.h>
+#include <plugin.h>
+#include <PluginUtilities.h>
+#include "Main.h"
+
+#include "../hookext_plugin/hookext_exports.h"
+
+map<string, uint> factions;
+map<string, float> factionReps;
```
dsyalex: this should be declared inside the UserCmd_Rep faction

Plugins/Public/getrep/Main.h
```diff
@@ -0,0 +1,20 @@
+#ifndef __MAIN_H__
```
dsyalex: I wonder if we can leave this file out - conn_plugin and hookext_plugin don't have it

PR #28, opened 2018-03-17
Compilation fixes: base.dll, npc.dll

ConradWeiser:
Namespace fix of the enum 'DestroyType' used within the base, and npc plugins.
With the fix applied, both plugins are able to be compiled again.

Ported commit from: [SVN Revision 284](https://forge.the-starport.net/projects/flhookplugin/repository/diff?utf8=%E2%9C%93&rev=284&rev_to=283) -- Only some of the lines were needed.
Credit to friendlyfire from the-starport

----

Commit revert occurred due to some work being done on the wrong branch. Only ID 79e8617 and 6562a70 are relevant.

### DIFF
--- a/Plugins/Public/alley/CustomPermissions.cpp
+++ b/Plugins/Public/alley/CustomPermissions.cpp
@@ -412,7 +412,7 @@ void ShowPlayerMissionText(uint iClientID, const wstring &text)
 	caption.begin_mad_lib(526999);
 	caption.end_mad_lib();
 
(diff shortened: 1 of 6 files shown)


zibbizor:
I'll leave this up for now but we're already in the process of updating this hook with the latest changes from FF.

dsyalex:
Yeah I think we're going to import all the TSP revisions rather than cherry-pick parts of random ones, so this may not get used, but thanks for the reminder. Leaving open until that's done.

dsyalex:
I still intend to do that, but let's use this for now.

PR #29, opened 2018-03-21
Dice extension

ConradWeiser:
There was a post on discoverygc saying that they'd like more dice features. 
Dice command rewritten to support 1d20[+-] operations.

Image attached for ingame visual

diceroll

The line explaining what each individual dice roll was is not displayed if there are more than 9 dice being rolled. Keep the screen clean from too much clutter. (and message limits somewhere down the line)

### DIFF
--- a/Plugins/Public/playercntl_plugin/Main.cpp
+++ b/Plugins/Public/playercntl_plugin/Main.cpp
@@ -903,7 +903,7 @@ USERCMD UserCmds[] =
 	{ L"/pos",			MiscCmds::UserCmd_Pos,			L"Usage: /pos" },
 	{ L"/stuck",		MiscCmds::UserCmd_Stuck,		L"Usage: /stuck" },
 	{ L"/droprep",		MiscCmds::UserCmd_DropRep,		L"Usage: /droprep" },
-	{ L"/dice",			MiscCmds::UserCmd_Dice,			L"Usage: /dice [max]" },
+	{ L"/roll",			MiscCmds::UserCmd_Dice,			L"Usage: /roll 1d20 | 1d20+3 | etc." },
 	{ L"/coin",			MiscCmds::UserCmd_Coin,			L"Usage: /coin" },
 	{ L"/pimpship",		PimpShip::UserCmd_PimpShip,		L"Usage: /pimpship" },
 	{ L"/showsetup",	PimpShip::UserCmd_ShowSetup,	L"Usage: /showsetup" },
@@ -1106,7 +1106,7 @@ void UserCmd_Help(uint iClientID, const wstring &wscParam)
 	PrintUserCmdText(iClientID, L"/pos");
 	PrintUserCmdText(iClientID, L"/stuck");
 	PrintUserCmdText(iClientID, L"/droprep");
-	PrintUserCmdText(iClientID, L"/dice [max]");
+	PrintUserCmdText(iClientID, L"/roll 1d20");
 	PrintUserCmdText(iClientID, L"/coin");
 
 	if (!set_bEnableRenameMe)
--- a/Plugins/Public/playercntl_plugin/Main.cpp
+++ b/Plugins/Public/playercntl_plugin/Main.cpp
@@ -903,6 +903,7 @@ USERCMD UserCmds[] =
 	{ L"/pos",			MiscCmds::UserCmd_Pos,			L"Usage: /pos" },
 	{ L"/stuck",		MiscCmds::UserCmd_Stuck,		L"Usage: /stuck" },
 	{ L"/droprep",		MiscCmds::UserCmd_DropRep,		L"Usage: /droprep" },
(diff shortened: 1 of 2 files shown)


ConradWeiser:
I suppose I could get rid of the enum, and simply shift things up into the lines which check to see if an operation exists in the regex smatch.

For readability however, I liked an enum.

dsyalex:
Can you link to the thread?

Plugins/Public/playercntl_plugin/Main.cpp
```diff
@@ -903,7 +903,7 @@ USERCMD UserCmds[] =
 	{ L"/pos",			MiscCmds::UserCmd_Pos,			L"Usage: /pos" },
 	{ L"/stuck",		MiscCmds::UserCmd_Stuck,		L"Usage: /stuck" },
 	{ L"/droprep",		MiscCmds::UserCmd_DropRep,		L"Usage: /droprep" },
-	{ L"/dice",			MiscCmds::UserCmd_Dice,			L"Usage: /dice [max]" },
```
dsyalex: Do we really want to remove existing commands?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -217,23 +226,99 @@ namespace MiscCmds
...
+			// Only print the steps taken if less than 10 dice was rolled.
+			if (rollCount < 10)
+			{
+				PrintLocalUserCmdText(iFromClientID, stows(diceResultSteps), set_iLocalChatRange);
+			}
+
+		}
+		else
+		{
+			PrintUserCmdText(iFromClientID, L"Usage: /roll 1d20");
```
dsyalex: Some comments about what each means would be nice

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -217,23 +226,99 @@ namespace MiscCmds
...
+			uint number = 0;
+
+			for (int i = 0; i < rollCount; i++)
+			{
+				int randValue = (rand() % diceCount) + 1;
+
+				// If we have a modifier, apply it
+				if (operation == diceOperation::ADD)
+				{
+					number += (randValue + modifierValue);
```
dsyalex: You can just always add randValue to number as soon as you've got it, drop line 286, turn this into number += modifierValue and turn line 281 into number -= modifierValue

ConradWeiser: I was aiming for consistency among all of the operations. Again. Readability. Instead of checking to see if the modifier value was positive or negative (thus requiring additional string parsing), I merely kept the modifier as a positive value. 

I suppose once I learned that the number was negative with the operation setter, the number could have been multiplied by -1, but the function overhead is quite non-existent. It didn't matter. Thus as a personal preference, it was set to what it is currently.

Plugins/Public/playercntl_plugin/Main.cpp
ConradWeiser: I mean. It could stay. But why, when the new one does the very same thing and more?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
ConradWeiser: Done.

```
PrintUserCmdText(iFromClientID, L"Usage: (NumDice) d (DiceSides) [+-] (Modifier)");
PrintUserCmdText(iFromClientID, L"Examples: /roll 1d20 -- Roll 1, 20 sided die");
PrintUserCmdText(iFromClientID, L"          /roll 1d8+4 -- Roll 1, 8 sided die and add 8");
PrintUserCmdText(iFromClientID, L"          /roll 4d20+2 -- Roll 4, 20 sided dice, adding 2 to each die rolled");
```

Plugins/Public/playercntl_plugin/Main.cpp
dsyalex: Can we not alias /dice to the new command?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: Nitpick: 'Roll 1, 8 sided dice, adding 8 to each die rolled' ?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: ok

Plugins/Public/playercntl_plugin/Main.cpp
ConradWeiser: A fair point. 
/dice has been re-added, and pointed to the same method.

/roll is being kept in, as that's the command generally used everywhere else on the internet for this sort of operation.

ConradWeiser:
Dice command has been aliased again.

PR #30, opened 2018-03-23
Remove the ability for ships to get stuck when blind jumping

LazDisco:
Created a new form of jump denial. death_system will kill the jumpship, but leave any ships brought. banned_system will cause the device to pick a different system instead. This keeps the exploration element of jumpdrives, while removing the ability for ships to jump into systems that they cannot escape from / ooRP systems.

### DIFF
--- a/Plugins/Public/playercntl_plugin/HyperJump.cpp
+++ b/Plugins/Public/playercntl_plugin/HyperJump.cpp
@@ -1876,8 +1876,8 @@ namespace HyperJump
 				systems.push_back(sysinfo->nickname);
 				sysinfo = Universe::GetNextSystem();
 			}
-			int x = 0;
-			while(x == 0)
+			bool x = true;
+			while(x)
 			{
 				// Pick a random system and position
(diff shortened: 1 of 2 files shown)


Plugins/Public/playercntl_plugin/HyperJump.cpp
```diff
@@ -1870,12 +1876,21 @@ namespace HyperJump
 				systems.push_back(sysinfo->nickname);
 				sysinfo = Universe::GetNextSystem();
 			}
-
-			// Pick a random system and position
-			jd.iTargetSystem = CreateID(systems[rand() % systems.size()].c_str());
-			jd.vTargetPosition.x = ((rand()*10) % 400000) - 200000.0f;
-			jd.vTargetPosition.y = ((rand()*10) % 400000) - 200000.0f;
-			jd.vTargetPosition.z = ((rand()*10) % 400000) - 200000.0f;
+			bool x = true;
```
dsyalex: bad variable naming

PR #31, opened 2018-03-25
/set dietime [on|off] command

ConradWeiser:
A toggle which timestamps death messages.

Screenshot

### DIFF
--- a/Plugins/Public/playercntl_plugin/Message.cpp
+++ b/Plugins/Public/playercntl_plugin/Message.cpp
@@ -1257,6 +1257,7 @@ namespace Message
 		{
 			PrintUserCmdText(iClientID, L"ERR Invalid parameters");
 			PrintUserCmdText(iClientID, usage);
+			return true;
 		}
 
 		wstring wscCharname = (const wchar_t*)Players.GetActiveCharacterName(iClientID);
--- a/Plugins/Public/playercntl_plugin/Message.cpp
+++ b/Plugins/Public/playercntl_plugin/Message.cpp
@@ -1478,8 +1478,6 @@ namespace Message
 				timeSent = true;
 			}
 
-			// Send for all players who currently has 
-
 			char *szXMLBuf;
 			int iXMLBufRet;
 			char *szXMLBufSys;
--- a/Plugins/Public/playercntl_plugin/Message.cpp
+++ b/Plugins/Public/playercntl_plugin/Message.cpp
@@ -201,6 +201,9 @@ namespace Message
 		// Current chat time settings
 		bool bShowChatTime;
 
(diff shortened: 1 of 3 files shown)


Plugins/Public/playercntl_plugin/Message.cpp
```diff
@@ -1241,6 +1245,34 @@ namespace Message
...
+		wstring wscParam1 = ToLower(GetParam(wscParam, ' ', 0));
+		bool bShowChatDieTime = false;
+		if (!wscParam1.compare(L"on"))
+			bShowChatDieTime = true;
+		else if (!wscParam1.compare(L"off"))
+			bShowChatDieTime = false;
+		else
+		{
+			PrintUserCmdText(iClientID, L"ERR Invalid parameters");
+			PrintUserCmdText(iClientID, usage);
```
dsyalex: return?

Plugins/Public/playercntl_plugin/Message.cpp
```diff
@@ -1241,6 +1245,34 @@ namespace Message
...
+		}
+
+		wstring wscCharname = (const wchar_t*)Players.GetActiveCharacterName(iClientID);
+
+		HookExt::IniSetB(iClientID, "msg.chat_dietime", bShowChatDieTime);
+
+		// Update the client cache.
+		map<uint, INFO>::iterator iter = mapInfo.find(iClientID);
+		if (iter != mapInfo.end())
+			iter->second.bShowChatDieTime = bShowChatDieTime;
```
dsyalex: Please can we just:
if (mapInfo.find(iClientID) != mapInfo.end())
    mapInfo[iClientID].bShowChatDieTime = bShowChatDieTime;
?

ConradWeiser: I've always been told that when working on a project which other people have started, try to use the same style of operations which the others have been using.

Reference line 1239
```// Update the client cache.
map<uint,INFO>::iterator iter=mapInfo.find(iClientID);
if (iter != mapInfo.end())
	iter->second.bShowChatTime = bShowChatTime;```

This was used in the UserCmd_SetChatTime method.
I'm simply mirroring what it used to complete pretty much the same task.

Plugins/Public/playercntl_plugin/Message.cpp
ConradWeiser: Whoops. I suppose we don't want to turn it off if invalid parameters were used. 
Fixed.

Plugins/Public/playercntl_plugin/Message.cpp
dsyalex: Good answer.

PR #33, opened 2018-04-05
Playercntl: Added tag based item restrictions

LazDisco:
Added tag based item restrictions. Functions the same way as items, but is dependent on the player's name. The idea behind this is too allow people to buy official faction ids without having to pester people to come online and give them an ID as all they need is the rename password.

### DIFF
--- a/Binaries/bin-conf/flhook_plugins/playercntl.cfg
+++ b/Binaries/bin-conf/flhook_plugins/playercntl.cfg
@@ -56,6 +56,7 @@ ItemMustBeMounted = yes
 [PurchaseRestrictions]
 EnforceIDRestrictions = yes
 CheckIDRestrictions = yes
+CheckTagRestrictions = yes
 
 [Message]
 SetMsg = yes
@@ -72,6 +73,9 @@ name = Player Control v1.0 by Cannon
 shortname = playercntl
 maypause = false
 mayunload = true
+
+[TagItemRestrictions]
+dsy_license_admin = [Admin]; The admin id can now only be got by anyone with the [Admin] in their name
 
 [dsy_license_br_n_grp]
 br_n_grp = 1.0, 3
--- a/Plugins/Public/playercntl_plugin/PurchaseRestrictions.cpp
+++ b/Plugins/Public/playercntl_plugin/PurchaseRestrictions.cpp
@@ -71,12 +71,12 @@ namespace PurchaseRestrictions
 	/// Log items of interest so we can see what cargo cheats people are using.
 	static void LogItemsOfInterest(uint iClientID, uint iGoodID, const string &details)
 	{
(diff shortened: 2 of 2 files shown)


ConradWeiser:
Variable and logic structures seem to be variants of what already exists, nothing appears to be missing from the logic itself, looks great!

The only recommendation I have, is that you include an example of what the configuration changes look like somewhere in the Binaries/bin-conf folder.

Neat little addition.

LazDisco:
Example done.

Binaries/bin-conf/flhook_plugins/playercntl.cfg
```diff
@@ -73,6 +74,9 @@ shortname = playercntl
 maypause = false
 mayunload = true
 
+[TagItemRestrictions]
+dsy_license_admin = [Admin]; The admin id can now only be got by anyone with the [Admin] in their name
```
dsyalex: Tags are only restricted at the beginning of names

LazDisco:
I've amended the code. Should now allow for multiple tags per item (split by comma), and I've tested it and can confirm that tags work no matter where in the name they are located.

dsyalex:
I think you misunderstand - tags are only restricted *as prefixes*, so it is not secure to base access controls on tags simply being contained within character names - it must be a prefix match. Please see playercntl Rename::CreateNewCharacter

Plugins/Public/playercntl_plugin/PurchaseRestrictions.cpp
```diff
@@ -236,6 +260,37 @@ namespace PurchaseRestrictions
 			}
 		}
 
+		// Restrict things via player tag
+		if (set_bCheckTagRestrictions)
+		{
+			for (map<uint, list<string>>::iterator iter = mapTagList.begin(); iter != mapTagList.end(); iter++)
+			{
+				if (iter->first == gbi.iGoodID)
```
dsyalex: direct lookup possible

Plugins/Public/playercntl_plugin/PurchaseRestrictions.cpp
```diff
@@ -236,6 +260,28 @@ namespace PurchaseRestrictions
...
 		}
 
+		// Restrict things via player tag
+		if (set_bCheckTagRestrictions)
+		{
+			if (mapTagList.find(gbi.iGoodID) != mapTagList.end())
+			{
+				list<string> tags = mapTagList[gbi.iGoodID];
+				string scCharname = wstos((const wchar_t*)Players.GetActiveCharacterName(iClientID));
+				bool validTag = find(tags.begin(), tags.end(), scCharname) != tags.end();
```
dsyalex: This appears to make it such that the check is not for tags, the actual whole character name must be in the config?

LazDisco:

PR #34, opened 2018-04-06
base_plugin: Add damage_tick_time config value for wear and tear tickrate

Kazinsal:
This change to the base plugin decouples the wear and tear tickrate from the general base tickrate. This will allow server admins to increase the time between wear and tear damage while not affecting the general tickrate of the base plugin.

This change adds the "damage_tick_time" configuration option to the base plugin's configuration file. It functions in the same manner as the "tick_time" configuration option, but for the wear and tear tickrate instead of the general tickrate.

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -472,7 +472,7 @@ bool CoreModule::Timer(uint time)
 
 			pub::SpaceObj::GetHealth(space_obj, base->base_health, base->max_base_health);
 
-			if (!dont_rust)
+			if (!dont_rust && ((time%set_damage_tick_time) != 0))
 			{
 				// Reduce hitpoints to reflect wear and tear. This will eventually
 				// destroy the base unless it is able to repair itself.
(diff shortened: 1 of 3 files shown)


Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -472,7 +472,7 @@ bool CoreModule::Timer(uint time)
 
 			pub::SpaceObj::GetHealth(space_obj, base->base_health, base->max_base_health);
 
-			if (!dont_rust)
+			if (!dont_rust && ((time%set_damage_tick_time) == 0))
```
dsyalex: I think we can get rid of the extra brackets

dsyalex:
I feel like we tried this before but okay.

Kazinsal:
My understanding is that the previous attempt involved no modification of FLHook and was literally "let's crank the tick time up to stop wear and tear oh wait that breaks every other thing about bases guess this is impossible without rewriting the base plugin".

Adding new features to this plugin is not impossible if you know where to do it.

dsyalex:
Let's give this a go anyway

PR #35, opened 2018-04-12
Add Message::SendChatEvent, eventmode logging for preset messages

Kazinsal:
Rebased to try to make two extra commits disappear but they didn't so who knows what's up with that, but they don't seem to actually exist in the PR's code itself so ???

Anyways what this does is adds a function to Message.cpp that logs its parameters as messages to the eventmode console, and calls it when sending preset messages (eg. /1, /s1, /t1, /g1) so they get picked up and logged by DSPM.

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -462,7 +462,7 @@ bool CoreModule::Timer(uint time)
 		return false;
 	}
 
-	if ((time%set_damage_tick_time) != 0)
+	if ((time%set_tick_time) != 0)
 		return false;
 
 	if (space_obj)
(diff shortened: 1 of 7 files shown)


dsyalex:
I would greatly prefer it if we could do this without duplicating half of HkCbIServerImpl.cpp's SubmitChat

Kazinsal:
We could move it all into FLHook itself and export it from global.h but then we're going to need to update the FLHook DLL on the server as well, which will probably complicate the deployment.

dsyalex:
Not really, it would just be another file to swap.

Kazinsal:
Changes made.

dsyalex:
This still isn't doing anything about SubmitChat in HkCbIServerImpl

PR #36, opened 2018-04-20
Nodock Plugin Reworked

LazDisco:
Under the new format, /nodock works by loading in different lists of bases, depending on the ID that initialised the nodock. Like the old plugin, police IDs can still restrict all civilian/military bases with nodock, as not to rob them of their main perk, but now each ID can block the bases of it's own IFF. So Outcasts can now nodock people to Outcast bases and Liberty Navy can nodock people to Liberty navy bases - and so on.

I've already filled out the entire of the config file and it's very easy to modify. In addition to the nodock function, it is possible to have certain IDs out right restricted from certain bases or solars. While I chose no to add any, it is an option. Helpful for events and RP developments. (Perhaps a non-hostile FR5 could be done with this)

Certain ships can be rendered exempt from all nodocks as shown inside of the config.


Binaries/bin-conf/flhook_plugins/laz_NoDock.cfg
```diff
@@ -0,0 +1,973 @@
...
+nodock = Li04_03
+nodock = LI14_01
+nodock = LI09_03
+nodock = LI12_02
+nodock = LI14_02
+nodock = LI13_02
+nodock = LI15_01
+nodock = LI15_01b
+;Guard Stations
+Li05_02
```
dsyalex: what's going on here?

Binaries/bin-conf/flhook_plugins/laz_NoDock.cfg
```diff
@@ -0,0 +1,973 @@
...
+nodock = HI18_02
+nodock = HI18_03
+nodock = HI20_01
+nodock = HI20_02
+nodock = HI20_03
+nodock = HI20_04
+nodock = RH11_01
+nodock = IW13_08
+;Guard Stations
+HI18_04
```
dsyalex: ?

Binaries/bin-conf/flhook_plugins/laz_NoDock.cfg
```diff
@@ -0,0 +1,973 @@
...
+nodock = Hi01_dock_ring_1
+nodock = Hi01_docking_fixture_1
+nodock = Hi01_02
+nodock = Hi01_03
+nodock = HI03_02
+nodock = HI05_01
+nodock = HI03_04
+nodock = GA06_02
+;Guard Stations
+HI10_01
```
dsyalex: ?

Plugins/Public/NoDock/Main.cpp
```diff
@@ -0,0 +1,574 @@
...
+{
+	returncode = DEFAULT_RETURNCODE;
+	int noDockIDS = 0; // How many IDs are we going to have
+	int noDockedBases = 0; // How many bases can be nodocked
+	int restrictedCount = 0; // How many solars are restricted
+	int ignoredShips = 0; // How many ships are ignored by nodock
+
+	char szCurDir[MAX_PATH];
+	GetCurrentDirectory(sizeof(szCurDir), szCurDir);
+	string NoDockCfg = string(szCurDir) + R"(\flhook_plugins\laz_NoDock.cfg)"; // Our Config File
```
dsyalex: waho what is going on here?

Plugins/Public/NoDock/Main.cpp
```diff
@@ -0,0 +1,574 @@
...
+			PrintUserCmdText(iClientID, L"The Nodock plugin is currently disabled.");
+		}
+		return true;
+	}
+
+	wstring wscCharacterName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(iClientID));
+	bool correctID = false;
+	uint noDocker; // The ID that preformed the nodock
+
+	for (auto& mapShip : mapShips)
```
dsyalex: this block looks like it could be turned into a direct lookup

Plugins/Public/NoDock/Main.cpp
```diff
@@ -0,0 +1,574 @@
...
+		const uint iClientIDTarget = HkGetClientIDByShip(iTarget);
+		if (!HkIsValidClientID(iClientIDTarget))
+		{
+			PrintUserCmdText(iClientID, L"ERR: Target is not a player.");
+			return true;
+		}
+
+		wstring wscTargetCharname = (const wchar_t*)Players.GetActiveCharacterName(iClientIDTarget);
+
+		for (auto& i : mapActiveNoDocks)
```
dsyalex: as above

Plugins/Public/NoDock/Main.cpp
```diff
@@ -0,0 +1,574 @@
...
+	HkEnumCargo((const wchar_t*)Players.GetActiveCharacterName(iClientID), lstCargo, iHoldSize);
+
+	for (auto ii : lstCargo)
+	{
+		auto* i = &ii;
+		if(i->bMounted)
+		{
+			for (auto& iter : mapIDRestricted)
+			{
+				if(iter.first == i->iArchID)
```
dsyalex: looks like another direct lookup case

Binaries/bin-conf/flhook_plugins/laz_NoDock.cfg
LazDisco: I was initially going to outright restrict guard stations so I left them out, but then I realised that might be a bad idea for RP, but forgot to go and add the nodock comments.

Plugins/Public/NoDock/Main.cpp
LazDisco: The entire thing is in a new project since you mentioned you desired projects to be separated rather than merged. This seemed like a good opportunity.

LazDisco:
The config was generated via a bit of code I wrote. It when through the game and extracted all to of the bases and their reps. I then used a regex expression to get only the bases with the reps of the faction I was currently adding to the config. 

For this reason, I haven't tested all the bases directly myself, but I am fairly sure the the code worked based on the few that I did test. I will use the script again to generate base names with it so we can comment all of the names.

Plugins/Public/NoDock/Main.cpp
dsyalex: Yeah my point was the raw string thing. I hadn't seen it before, but it's okay.

dsyalex:
I think I'll need a copy of that code to review and run myself to generate the actual config

Binaries/bin-conf/flhook_plugins/laz_NoDock.cfg
dsyalex: Those are keys but okay

zibbizor:
If you are using external tools to generate config files it may be a good time for us to create a new folder (or repo?) to house these projects. The plugins are a bit useless if they're a config nightmare like this is.

It may be time to consider using a naming convention as well.

LazDisco:
https://github.com/DiscoveryGC/Tools/blob/master/Config%20Creator/NoDock/Program.cs

I wrote a program to generate it, and moved it to a shared repo. This tool isn't 100% accurate due to how Discovery does stuff, but it gets you most of the way there.

dsyalex:
So are you saying that it won't actually produce this config?

LazDisco:
> So are you saying that it won't actually produce this config?

I have just updated the generation utility to be significantly more accurate and do more of the leg work in actually generating it. There are a few factions still being left out for reasons I am trying to figure out; HF, AI, Coalition, are good examples.

The config I have just updated is almost exactly like the one the tool generates and is server usable as it stands. I'll work on covering a few of the left out factions, but for now it's a minor amount of manual editing.

LazDisco:

PR #63, opened 2018-04-30
Playercntl modifications

LazDisco:
Added optional reasons for givecash/givecasht
Added the ability for Admins to use ".jumpgroup" to jump everyone within a 1k radius to a specific point in a specified system.
Added ".pos" for admins that gives them the current pos value, as well as the internal id of what ever system they are in.
Change /pos to include the current sector.
Split the pimpship information into it's own command.
Added the ability for people to swap out the hardpoint ID for the word "all" which will change all hardpoints to the specified light.
Added the ability for people to swap out the light ID for the name of the desired light.


LazDisco:
Will redo with proper commit history.

PR #73, opened 2018-07-11
Re-add stackables to autobuy plugin

Kazinsal:
This commit re-adds stackable ammo counts for missiles and guns to the autobuy plugin. The list of weapons that are stackable is configured in `autobuy.cfg`, with the following format:

```
[stackable]
weapon = torpedo03_mark01
weapon = dsy_silencer_missile
weapon = dsy_nightmare_torpedo
```

Any gun, missile launcher, or torpedo launcher given a `weapon = <id>` entry under the `[stackable]` section in `autobuy.cfg` will be considered stackable, ie. having multiple of the same launcher equipped will increase the amount of ammo the autobuy plugin loads into your ship accordingly. For example, equipping two Hellfire Rocket Pods will result in twice the ammo capacity of a single rocket pod (120 each) being loaded into your ship by the autobuy plugin.

### DIFF
--- a/Plugins/Public/autobuy/Main.cpp
+++ b/Plugins/Public/autobuy/Main.cpp
@@ -102,6 +102,7 @@ void LoadSettings()
 	string File_FLHook = "..\\exe\\flhook_plugins\\autobuy.cfg";
 	int iLoaded = 0;
 	int iLoaded2 = 0;
+	int iLoadedStackables = 0;
 
 	INI_Reader ini;
 	if (ini.open(File_Misc.c_str(), false))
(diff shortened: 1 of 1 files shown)


Plugins/Public/autobuy/Main.cpp
```diff
@@ -220,18 +221,25 @@ void LoadSettings()
 						}
 					}
 				}
+				else if (ini.is_header("stackable"))
+				{
+					while (ini.read_value())
+					{
+						if (ini.is_value("weapon"))
+						{
+							mapStackableItems[CreateID(ini.get_value_string(0))] = true;
```
dsyalex: shouldn't this be a list/set ?

Kazinsal:
It's a map in the original implementation. I believe the intent was to allow partial stacking down the line if the balance team wanted that. Not 100 percent sure because it was a quick hack in the first place.

zibbizor:
Yeah, the initial beta testing was done on hellfire pods to test the feature and its balance implications. It was pretty much left like that for a while as Haste went inactive and no progress was made.

PR #74, opened 2018-08-27
Solution cleaning

LazDisco:
Moved a lot of useful functions from various plugins into PluginUtilities where they should belong. Cleaned up quite a bit of reused and duplicated code, but I have no doubt there is more. I will make additional changes as I find them.

The most unusual functions I've moved from here will probably be MailSend. I personally find it to be a really useful utility for delivering information to offline people and have used it in a number of plugins so far. The rest of the mail functions have stayed where they belong, but I believe MailSend is significantly more useful being accessible by other plugins.


Plugins/Public/playercntl_plugin/Mail.h
```diff
@@ -13,7 +13,6 @@ namespace Mail
 	void MailShow(const wstring &wscCharname, const string &scExtension, int iFirstMsg);
 	bool MailDel(const wstring &wscCharname, const string &scExtension, int iMsg);
 	void MailCheckLog(const wstring &wscCharname, const string &scExtension);
-	bool MailSend(const wstring &wscCharname, const string &scExtension, const wstring &wscMsg);
```
dsyalex: I think this one should stay as-is.

Plugins/Public/base_plugin/Main.h
```diff
@@ -576,8 +576,6 @@ extern float set_shield_damage_multiplier;
 /// Holiday mode
 extern bool set_holiday_mode;
 
-wstring HtmlEncode(wstring text);
```
dsyalex: This one seems suspect to me. It likely isn't appropriate for all use cases.

Plugins/Public/playercntl_plugin/Mail.h
LazDisco: I understand why, but I personally think the mail plugin should be part of FLHook Core due to how useful it is as a utility. Perhaps this is something we should consider porting out of Playercntl entirely and moving into Core. 

The ability to send messages while offline is something we'll use again, and I feel I can say that with a degree of certainty. It'll be problematic if we have to integrate a new solution each and every time we do it.

Plugins/Public/base_plugin/Main.h
LazDisco: I looked at the function, I don't see anything _glaringly_ off about it.

LazDisco:

Issue #48, opened 2018-04-29
Check notifications when miner is full


dsyalex:
Does the Cargo Full sound get played on both ends? What about also adding a text notification?

dsyalex:
Plugins/Public/event/Main.cpp:                  pub::Player::SendNNMessage(iClientID, CreateID("insufficient_cargo_space"));
Plugins/Public/minecontrol_plugin/Main.cpp:                                                     pub::Player::SendNNMessage(iClientID, CreateID("insufficient_cargo_space"));

Looks like it's only played on the miner's end - also send to target transport. Should add a text notification too.

dsyalex:
Fixed by #137

PR #76, opened 2018-08-27
Miner notifications

LazDisco:
Added text based notifications to miners when transport's cargo is full, and also made it so transports get both an audio and a text notification.

#48

### DIFF
--- a/Plugins/Public/event/Main.cpp
+++ b/Plugins/Public/event/Main.cpp
@@ -914,10 +914,13 @@ void __stdcall SPMunitionCollision(struct SSPMunitionCollisionInfo const & ci, u
 		}
 		if (iLootCount == 0)
 		{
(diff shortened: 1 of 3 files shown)


Plugins/Public/event/Main.cpp
```diff
@@ -914,7 +914,10 @@ void __stdcall SPMunitionCollision(struct SSPMunitionCollisionInfo const & ci, u
 		}
 		if (iLootCount == 0)
 		{
+			PrintUserCmdText(iClientID, L"%s's cargo is now full.", reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(iSendToClientID)));
```
dsyalex: we don't want to run this or the NNMessage if iClientID == iSendToClientID (same in minecontrol). I suggest you make a if (iClientID != iSendToClientID) block and move the stuff involving iSendToClientID into that.

LazDisco:

PR #80, opened 2018-10-11
Base Crew Fix

LazDisco:
As discussed in the dev subforum, this change allows the crew commodity to be ignored, with the FOW requirements still in effect and scaling as intended.

I did do a bit of testing and the change seems to work as intended when tested with a basic Core 1 POB. Base will heal without crew, but only when all the materials, in the right amounts+, are present.


Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -408,7 +408,7 @@ void CoreModule::RepairDamage(float max_base_health)
 	{
 		// Use water and oxygen.
 		uint ow_available = base->HasMarketItem(i->first);
-		if (ow_available >= 250)
+		if (ow_available >= 250 * base->base_level)
```
dsyalex: what does this have to do with this commit?

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -420,7 +420,7 @@ void CoreModule::RepairDamage(float max_base_health)
 	i != set_base_crew_food_items.end(); ++i)
 	{
 		uint food_available = base->HasMarketItem(i->first);
-		if (food_available >= 250)
+		if (food_available >= 250 * base->base_level)
```
dsyalex: ^

Plugins/Public/base_plugin/Main.h
```diff
@@ -576,7 +576,8 @@ extern float set_shield_damage_multiplier;
 /// Holiday mode
 extern bool set_holiday_mode;
 
-wstring HtmlEncode(wstring text);
```
dsyalex: why is this being removed?

LazDisco:

PR #83, opened 2018-11-04
Zone Utils

LazDisco:
Playercntl / MiningControl contained some very useful functions for working with zones, something FLHook Core lacks. I've merged these two files together (since they were slightly different) and given them their own place in PluginUtilities. I did some testing and this version *appears* to work fine with all plugins.


LazDisco:

PR #88, opened 2019-06-13
dock v0.9.2

lnvoker:
My attempt to fix docking modules.

### DIFF
--- a/Plugins/Public/mobiledocking_plugin/Main.cpp
+++ b/Plugins/Public/mobiledocking_plugin/Main.cpp
@@ -3,119 +3,323 @@
  * Initial Version by Cannon
  * Using some work written by Alley
  * Rework by Remnant
+ * Resurrected by Invoker
  */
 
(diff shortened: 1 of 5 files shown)


zibbizor:
What in the world is going on here? 

- #pragma once does the job, you don't need defs.

 - Why are you using threads like this? This is both the weirdest and most dangerous thing I've seen submitted here.

 - "// This plugin always runs before the Conn Plugin runs it's /conn function. Verify that there are no docked ships." NO absolutely not. You seem to have great misconceptions on how FLHook works. The execution order for the same hook is load order based, which means the conn plugin will execute before something called dock plugin. Reloads are the exception to this.

What you are doing is a very different design than the original. It's not good to overwrite the existing mobiledock plugin. Make a new one with the plugin template

lnvoker:
Took all this into account.
Gonna rewrite some stuff tomorrow.

PR #90, opened 2019-06-14
Docking modules plugin v0.4.3

lnvoker:
This PR supposed to be bug fix only. Main changes:
1) Forced system switch removed completely. I think that is cause of most nasty bug.
2) SaveFiles.cpp removed, this part didn't its job somehow properly anyway.


Plugins/Public/mobiledocking_plugin/Main.h
```diff
@@ -38,18 +41,26 @@ struct CLIENT_DATA
...
+	wstring dockedCharname;
 };
 
-static map<uint, DEFERREDJUMPS> mapDeferredJumps;
+extern vector<ActionJettison> jettisonList;
 
-void LoadShip(string shipFileName);
-void SaveDockInfoCarrier(const wstring& shipFileName, uint clientID, const CLIENT_DATA& client);
-void SaveDockInfoCarried(const wstring& shipFileName, uint clientID, const CLIENT_DATA& client);
+struct DelayedJettison
```
dsyalex: why is this function in a struct?

Plugins/Public/mobiledocking_plugin/MobileDocking Plugin.vc14.vcxproj
```diff
@@ -132,7 +132,6 @@ copy /Y $(OutDir)$(TargetName).pdb  ..\..\..\Binaries\bin-vc14\flhook_plugins\pu
   <ItemGroup>
     <ClCompile Include="ClientCommands.cpp" />
     <ClCompile Include="Main.cpp" />
-    <ClCompile Include="SaveFiles.cpp" />
```
dsyalex: I think we should list the pros and cons of removing that functionality.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -194,6 +229,9 @@ void __stdcall BaseExit(uint iBaseID, uint iClientID)
 
 }
 
+// Temporary storage for client data to be handled in LaunchPosHook.
+CLIENT_DATA undockingShip;
```
dsyalex: note to self: this is done in the base plugin already. Set in PlayerLaunch, read in LaunchPosHook

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,57 @@ vector<uint> dockingModuleEquipmentIds;
...
-	mapDeferredJumps[client].rot = ornt;
+	const wchar_t* carrierCharname = (const wchar_t*)Players.GetActiveCharacterName(carrierClientID);
+	const wchar_t* dockedCharname = (const wchar_t*)Players.GetActiveCharacterName(dockedClientID);
 
-	// Send the jump command to the client. The client will send a system switch out complete
-	// event which we intercept to set the new starting positions.
-	PrintUserCmdText(client, L" ChangeSys %u", system);
+	if (dockedClientID != -1)
+	{
+		PrintUserCmdText(carrierClientID, L"Ship warned. If it won't undock in %i seconds, it will be kicked by force.", jettisonKickTime);
```
dsyalex: s/won't/doesn't/

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -306,6 +380,22 @@ void __stdcall PlayerLaunch_AFTER(unsigned int ship, unsigned int client)
...
 	}
 
+	// Land ship to proxy base in carrier's system if they are in different systems.
+	if (mobiledockClients[client].carrierSystem == -1)
+	{
+		string scProxyBase = HkGetPlayerSystemS(HkGetClientIdFromCharname(mobiledockClients[client].wscDockedWithCharname)) + "_proxy_base";
+		uint iBaseID;
+		if (pub::GetBaseID(iBaseID, scProxyBase.c_str()) == -4)
+		{
+			PrintUserCmdText(client, L"No proxy base in system detected. Contact a developer about this please.");
```
dsyalex: add a thing about telling the dev which system they tried to undock in

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -306,6 +380,22 @@ void __stdcall PlayerLaunch_AFTER(unsigned int ship, unsigned int client)
...
+			return;
+		}
+
+		pub::Player::ForceLand(client, iBaseID);
+		PrintUserCmdText(client, L"Navmap updated succesfully.");
```
dsyalex: seems like a bizarre message?

LazDisco:
You don't need to close a pull request every time you change something. It would probably be worth taking on board what Alley wrote [here](https://github.com/DiscoveryGC/FLHook/pull/88). Additionally, to add to that, and as I have already mentioned to you over PM, threading inside of FLServer if you don't know what you are doing is very dangerous.

There is a reason it was never added to Core FLHook despite C++ being able to do it naively since 2011, not to mention that we also have boost included which has support for threading as well.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,57 @@ vector<uint> dockingModuleEquipmentIds;
...
+			uint checkCarrierClientID = HkGetClientIdFromCharname(it->carrierCharname);
+			uint checkDockedClientID = HkGetClientIdFromCharname(it->dockedCharname);
+
+			// If both carrier and docked ship are still at server.
+			if (checkDockedClientID != -1 && checkCarrierClientID != -1)
+			{
+				if (mobiledockClients[checkDockedClientID].wscDockedWithCharname == it->carrierCharname)
+				{
+					mobiledockClients[checkCarrierClientID].mapDockedShips.erase(it->dockedCharname);
+					HkKick(HkGetAccountByClientID(checkDockedClientID));
```
dsyalex: HkKickReason with ARG_CLIENTID

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -235,30 +268,55 @@ void __stdcall PlayerLaunch(unsigned int iShip, unsigned int client)
...
-		// Clear the client out of the mobiledockClients now that it's no longer docked
-		mobiledockClients.erase(client);
+			// Now that the data is prepared, send the player to the carrier's location
+			undockingShip = mobiledockClients[client];
+
+			// Clear the client out of the mobiledockClients now that it's no longer docked
+			mobiledockClients.erase(client);
+
+			// Remove charname from carrier's mapDockedShips.
+			mobiledockClients[carrier_client].mapDockedShips.erase(clientName);
```
dsyalex: iDockingModulesAvailable++?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -235,30 +268,55 @@ void __stdcall PlayerLaunch(unsigned int iShip, unsigned int client)
...
+			if (iBaseID)
 			{
-				return;
+				// Set the flags which the PlayerLaunch_AFTER uses to handle teleporting to the carriers base
+				mobiledockClients[client].undockBase = Universe::get_base(iBaseID);
+				mobiledockClients[client].baseUndock = true;
 			}
+			else
+			{
+				// If died carrier and docked ship are in same system.
```
dsyalex: died carrier?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -375,23 +465,6 @@ void __stdcall BaseEnter(uint iBaseID, uint client)
 		status += L"<TEXT>" + XMLText(mobiledockClients[client].wscDockedWithCharname) + L"</TEXT><PARA/><PARA/>";
 		status += L"<POP/></RDL>";
 		SendSetBaseInfoText2(client, status);
-
-		// Check to see that the carrier thinks this ship is docked to it.
-		// If it isn't then eject the ship to space.
```
dsyalex: any idea how this was possible before? looks like IsShipDockedOnCarrier function is no longer in use

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: If ship doesn't exists and client is not at base, then probably client is watching death screen.
May bug up if carrier respawns before docked ships are launched in space.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Possible with help of SaveFiles.cpp.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: It is HkKick() and it works fine without ARG_CLIENTID

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Not more bizarre if player tries to undock and sees that he is still at proxy base without any explanation.

Plugins/Public/mobiledocking_plugin/Main.h
lnvoker: I'd like to think that I create an object.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Fixed.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Now it tells to player which base is required.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,57 @@ vector<uint> dockingModuleEquipmentIds;
...
+	const wchar_t* carrierCharname = (const wchar_t*)Players.GetActiveCharacterName(carrierClientID);
+	const wchar_t* dockedCharname = (const wchar_t*)Players.GetActiveCharacterName(dockedClientID);
 
-	// Send the jump command to the client. The client will send a system switch out complete
-	// event which we intercept to set the new starting positions.
-	PrintUserCmdText(client, L" ChangeSys %u", system);
+	if (dockedClientID != -1)
+	{
+		PrintUserCmdText(carrierClientID, L"Ship warned. If it won't undock in %i seconds, it will be kicked by force.", jettisonKickTime);
+		PrintUserCmdText(dockedClientID, L"Carrier wants to jettison your ship. Undock willingly or you will be kicked after %i seconds.", jettisonKickTime);
```
dsyalex: or carrier has died?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -522,13 +591,10 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
 
 			//Force the docked ship to launch. The teleport coordinates have been set by the previous method
 			const CLIENT_DATA& dockedShipData = mobiledockClients[iDockedClientID];
-			JumpToLocation(iDockedClientID, dockedShipData.carrierSystem, dockedShipData.carrierPos, dockedShipData.carrierRot);
+			mobiledockClients[client].iDockingModulesAvailable++;
```
dsyalex: this should probably happen when the ship actually gets jettisoned, rather than immediately

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -614,64 +681,32 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
 	return false;
 }
 
-void __stdcall JumpInComplete(uint iSystemID, uint iShip)
```
dsyalex: so wait how do we get away without this one?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -614,64 +681,32 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
+		// If carrier is present at server - do it, if not - whatever. Plugin erases all associated client data after disconnect. 
+		if (carrierClientID != -1)
 		{
-			wstring shipFileName;
-			wstring charName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(iClientID));
-
-			HkGetCharFileName(charName, shipFileName);
-			SaveDockInfoCarried(shipFileName, iClientID, mobiledockClients[iClientID]);
+			wstring charname = (const wchar_t*)Players.GetActiveCharacterName(iClientID);
+			mobiledockClients[carrierClientID].mapDockedShips.erase(charname);
```
dsyalex: iDockingModulesAvailable++?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -614,64 +681,32 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
 		}
 
-		mobiledockClients.erase(iClientID);
+		// Redirect client to last real base.
```
dsyalex: what if they last docked on a POB?

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: Yes but we should give a reason when kicking

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: It compares current systems from PlayersDB.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -134,8 +169,8 @@ void LoadSettings()
 	CreateDirectoryA(moddir.c_str(), 0);
 
 	// Plugin configuration
-	char szCurDir[MAX_PATH];
-	GetCurrentDirectory(sizeof(szCurDir), szCurDir);
+	char szCurDir[MAX_PATH];
+	GetCurrentDirectory(sizeof(szCurDir), szCurDir);
```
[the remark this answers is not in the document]
Kazinsal: Might be line ending conversions

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -306,6 +380,22 @@ void __stdcall PlayerLaunch_AFTER(unsigned int ship, unsigned int client)
 		mobiledockClients.erase(client);
 	}
 
+	// Land ship to proxy base in carrier's system if they are in different systems.
+	if (mobiledockClients[client].carrierSystem == -1)
+	{
+		string scProxyBase = HkGetPlayerSystemS(HkGetClientIdFromCharname(mobiledockClients[client].wscDockedWithCharname)) + "_proxy_base";
+		uint iBaseID;
+		if (pub::GetBaseID(iBaseID, scProxyBase.c_str()) == -4)
```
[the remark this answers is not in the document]
Kazinsal: That magic constant as a return value for pub::GetBaseID() seems to be persistent throughout FLHook itself. No idea what it actually should be defined as but it seems to mean "base doesn't exist"

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Fixed(I hope).

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Now it shows reason.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Fixed.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Same messages in cases if carrier died or wants to jettison ship - is not a problem.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Fixed.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: They will be redirected to proxy base in same system with the POB.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Fixed.

zibbizor:
As started in the other PR, please do not overwrite the existing plugin. Make a new one.

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: No, this is a problem. When the carrier dies the message docked clients are told will be a lie. The message should read, based on the current callers, 'Either the carrier wants to jettison your ship, or it has died.'

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -62,27 +105,14 @@ void LogCheater(uint client, const wstring &reason)
...
-void UpdateCarrierLocationInformation(uint dockedClientId, uint carrierShip) 
+void UpdateCarrierLocationInformation(uint dockedClientId, uint carrierShip)
 {
 
 	// Prepare references to the docked ship's copy of the carrierShip's position and rotation for manipulation
 	Vector& carrierPos = mobiledockClients[dockedClientId].carrierPos;
 	Matrix& carrierRot = mobiledockClients[dockedClientId].carrierRot;
 
-		
+
```
dsyalex: whitespace change, out of scope for this PR

Plugins/Public/mobiledocking_plugin/Main.cpp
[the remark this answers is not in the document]
dsyalex: this PR is not for making line ending conversions in. If you want to do that, please make a separate PR

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -134,26 +163,26 @@ void LoadSettings()
...
+				while (ini.read_value())
 				{
-					if(ini.is_value("allowedmodule"))
+					if (ini.is_value("allowedmodule"))
 					{
 						dockingModuleEquipmentIds.push_back(CreateID(ini.get_value_string()));
 						dockingModAmount++;
 					}
-					else if(ini.is_value("cargo_capacity_limit"))
+					else if (ini.is_value("cargo_capacity_limit"))
```
dsyalex: lots more irrelevant whitespace changes down to here

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -171,12 +200,14 @@ void __stdcall BaseExit(uint iBaseID, uint iClientID)
 {
 	returncode = DEFAULT_RETURNCODE;
 
+	mobiledockClients[iClientID].iDockingModulesInstalled = 0;
+
 	// Check to see if the vessel undocking currently has a docking module equipped
-	for(list<EquipDesc>::iterator item = Players[iClientID].equipDescList.equip.begin(); item != Players[iClientID].equipDescList.equip.end(); item++)
```
dsyalex: more irrelevant whitespace changes from here down

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -205,26 +237,47 @@ void __stdcall PlayerLaunch(unsigned int iShip, unsigned int client)
...
-		else
+		else if (carrier_client == -1)
 		{
 			// If the carrier doesn't exist for some reason, force the client to dock on it's last known base
 			uint iBaseID = mobiledockClients[client].iLastBaseID;
 			mobiledockClients[client].undockBase = Universe::get_base(iBaseID);
 			mobiledockClients[client].baseUndock = true;
 
 			return;
-
```
dsyalex: why are we doing this stuff in this PR

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -458,36 +519,36 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
 {
 	returncode = DEFAULT_RETURNCODE;
 
-	if(wscCmd.find(L"/listdocked") == 0)
```
dsyalex: more whitespace changes from here

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -496,43 +557,35 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
 		}
 
-		mobiledockClients[client].iDockingModulesAvailable++;
-		mobiledockClients.erase(iDockedClientID);
-		mobiledockClients[client].mapDockedShips.erase(charname);
-		PrintUserCmdText(client, L"Ship jettisoned");
 		return true;
 
 	}
-	else if(wscCmd.find(L"/allowdock")==0)
```
dsyalex: more

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: To me it seems that we don't actually keep it up to date all the time though? Is that okay?

dsyalex:
I think this PR is capable of being a patch to the existing code rather than a complete rewrite

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,64 @@ vector<uint> dockingModuleEquipmentIds;
...
+			if (checkDockedClientID != -1 && checkCarrierClientID != -1)
+			{
+				// If the client is still docked
+				if (mobiledockClients[checkDockedClientID].mobileDocked)
+				{
+					HkKickReason(ARG_CLIENTID(checkDockedClientID), L"Forced undocking.");
+
+					// If wscDockedWithCharname, that does mean that carrier died and next message is not required.
+					if(!mobiledockClients[checkDockedClientID].wscDockedWithCharname.empty())
+					PrintUserCmdText(checkCarrierClientID, L"Jettisoning ship...");
```
dsyalex: indent even if youre using an if without braces

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,64 @@ vector<uint> dockingModuleEquipmentIds;
...
+			uint checkCarrierClientID = HkGetClientIdFromCharname(it->carrierCharname);
+			uint checkDockedClientID = HkGetClientIdFromCharname(it->dockedCharname);
+
+			// If both carrier and docked ship are still at server.
+			if (checkDockedClientID != -1 && checkCarrierClientID != -1)
+			{
+				// If the client is still docked
+				if (mobiledockClients[checkDockedClientID].mobileDocked)
+				{
+					HkKickReason(ARG_CLIENTID(checkDockedClientID), L"Forced undocking.");
```
dsyalex: why are we no longer doing mapDockedShips here?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -295,17 +358,50 @@ void __stdcall PlayerLaunch(unsigned int iShip, unsigned int client)
...
+			string scProxyBase = HkGetPlayerSystemS(HkGetClientIdFromCharname(mobiledockClients[client].wscDockedWithCharname)) + "_proxy_base";
+			if (pub::GetBaseID(mobiledockClients[client].proxyBaseID, scProxyBase.c_str()) == -4)
+			{
+				PrintUserCmdText(client, L"No proxy base in system detected. Contact a developer about this please. Required base: %s", scProxyBase);
+				return;
+			}
+		}
+
+		pub::Player::ForceLand(client, mobiledockClients[client].proxyBaseID);
+		PrintUserCmdText(client, L"Navmap updated successfully.");
```
dsyalex: this message makes no sense?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -416,39 +494,22 @@ void __stdcall ShipDestroyed(DamageList *_dmg, DWORD *ecx, uint kill)
...
-					if(!mobiledockClients[iDockedClientID].mobileDocked)
-					{
-						mobiledockClients[client].mapDockedShips.erase(i->first);
-						continue;
-					}
-
 					// Update the coordinates the given ship should launch to.
 					UpdateCarrierLocationInformation(iDockedClientID, cship->get_position(), cship->get_orientation());
 
+					// Carrier is no more. Remove character name.
```
dsyalex: this feels like a weird hack

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -205,26 +237,47 @@ void __stdcall PlayerLaunch(unsigned int iShip, unsigned int client)
 
 	if (mobiledockClients[client].mobileDocked)
 	{
-
 		returncode = SKIPPLUGINS;
 
-		// Update the location of the carrier and remove the docked ship from the carrier
-		if (carrier_client != -1)
+		// Set last base to last real base this ship was on. If this was POB... well, base plugin should export more functions to work with.
```
dsyalex: I think it does now - see the latest commit to the repo, regarding GetCustomLastBaseForClient/CUSTOM_BASE_LAST_DOCKED

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: It is being done automatically on DisConnect()

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Tests show that it is okay.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,64 @@ vector<uint> dockingModuleEquipmentIds;
...
+					if(!mobiledockClients[checkDockedClientID].wscDockedWithCharname.empty())
+					PrintUserCmdText(checkCarrierClientID, L"Jettisoning ship...");
+				}
+			}
+
+			jettisonList.erase(it);
+		}
+		else
+		{
+			it++;
```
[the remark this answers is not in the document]
lnvoker: It really makes if one of elements in vector was erased DURING iterating through it in for loop.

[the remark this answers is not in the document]
Kazinsal: Did you test this change? It seems incredibly dangerous.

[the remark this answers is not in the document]
lnvoker: I did tests before. Server crashed until I removed automatic ++ of iterator and made it to be done only if no elements were erased from vector.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,64 @@ vector<uint> dockingModuleEquipmentIds;
...
+		if (it->timeLeft == 0)
+		{
+			uint checkCarrierClientID = HkGetClientIdFromCharname(it->carrierCharname);
+			uint checkDockedClientID = HkGetClientIdFromCharname(it->dockedCharname);
+
+			// If both carrier and docked ship are still at server.
+			if (checkDockedClientID != -1 && checkCarrierClientID != -1)
+			{
+				// If the client is still docked
+				if (mobiledockClients[checkDockedClientID].mobileDocked)
```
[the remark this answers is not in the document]
lnvoker: Solved.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Fixed.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Makes with added commentary.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Now plugin uses flag to check if ship docked on dead carrier.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Solved.

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: Invoker PM'd me: 'When carrier goes to another system, docked ships remain in previous with outdated system navigational map.'
Resolving

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: Let's make this into a TODO to implement support for at a later stage.

Plugins/Public/mobiledocking_plugin/Main.cpp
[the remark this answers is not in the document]
dsyalex: What if we start from the end and go through the list backwards with --, removing items along the way? This code appears wrong as-is.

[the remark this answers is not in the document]
lnvoker: The code works well, but there seems to be better alternative. Should test that later.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -15,21 +15,71 @@ vector<uint> dockingModuleEquipmentIds;
...
+		if (!mobiledockClients[dockedClientID].carrierDied)
+		{
+			PrintUserCmdText(carrierClientID, L"Ship warned. If it doesn't undock in %i seconds, it will be kicked by force.", jettisonKickTime);
+			PrintUserCmdText(dockedClientID, L"Carrier wants to jettison your ship. Undock willingly or you will be kicked after %i seconds.", jettisonKickTime);
+		}
+		else
+		{
+			PrintUserCmdText(dockedClientID, L"ALERT! Carrier ship is being destroyed. Launch ship or you will die in explosion.");
+		}
+		pub::Audio::PlaySoundEffect(dockedClientID, CreateID("rtc_klaxon_loop"));
```
LazDisco: I don't think we should hardcode stuff like this. I get it was like this in the old plugin, but the old plugin was made with enough bad design decisions as it stands. I'd suggest setting it up as a config value that can be changed.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -614,64 +681,30 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
-	
-	// Attempt to load the data for this ship if it exists. Trim off the .fl at the end of the file name.
-	string charFilename = cId.szCharFilename;
-	charFilename = charFilename.substr(0, charFilename.size() - 3);
-
-	LoadShip(charFilename);
-
-}
-
 void __stdcall DisConnect(uint iClientID, enum EFLConnection p2)
```
LazDisco: While you can use DisConnect for this, the convention is to handle the cleaning of client data on the connection of a new client. That's why ClearClientInfo exists.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Nice idea. Will be done in full version of the plugin.

Plugins/Public/mobiledocking_plugin/Main.cpp
lnvoker: Makes sense. Removed.

Plugins/Public/mobiledocking_plugin/Main.cpp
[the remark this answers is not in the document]
dsyalex: Instead of discarding the result of `jettisonList.erase(it);`, use it like this:
`it = jettisonList.erase(it);`
https://en.cppreference.com/w/cpp/container/vector/erase
Then this will all make sense I think.

Plugins/Public/mobiledocking_plugin/Main.cpp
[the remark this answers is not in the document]
dsyalex: (me nitpicking, will not really affect status of PR)

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -295,6 +353,22 @@ void __stdcall PlayerLaunch(unsigned int iShip, unsigned int client)
 	}
 }
 
+bool __stdcall LaunchPosHook(uint space_obj, struct CEqObj &p1, Vector &pos, Matrix &rot, int dock_mode)
+{
+	returncode = DEFAULT_RETURNCODE;
+
+	// Redirect the ship to carrier's position. Can bug with POB plugin, changes may be required.
```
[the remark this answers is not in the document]
dsyalex: paging @Kazinsal

Plugins/Public/mobiledocking_plugin/Main.h
dsyalex: Why?

Plugins/Public/mobiledocking_plugin/MobileDocking Plugin.vc14.vcxproj
dsyalex: We sort of discussed this in PM but I'd like to see a more formal justification laid out on the PR

lnvoker: About why I removed SaveFiles.cpp.
1) That thing is built on client ID to assign loaded data from file to that ID, what is absolutely totally wrong, because client ID is temporary variable. And when player enters the server next time, it may be (actually in 99% cases) not same.
2) Serializing seems to be not complete, it does its job well in half of cases, if turn blind eye on 1), but other half will generate bugs(plugin tells you that you have some ships on board, when actually not), which seem to be critical with new jettisoning mechanic. It removes ship only AFTER player was kicked and player remains docked to carrier until being kicked. Carrier is not able to jettison ship, which is not docked to him.

Plugins/Public/mobiledocking_plugin/Main.cpp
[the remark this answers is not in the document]
lnvoker: Done.

Plugins/Public/mobiledocking_plugin/Main.cpp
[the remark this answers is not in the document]
lnvoker: I set priority for LaunchPosHook to -1, so POB plugin will always handle it first, setting returncode to SKIPPLUGINS_NOFUNCTIONCALL, what prevent docking modules from handling LaunchPosHook after base plugin did.

Plugins/Public/mobiledocking_plugin/Main.h
lnvoker: Really, why? Now function is extracted from struct.

dsyalex:
This is looking promising and I'm reasonably happy with the code now. I have just been playing around with it locally. I've found that if you get kicked for refusing to undock on jettisonship, you respawn on a proxy base... I'm not sure there's a way around that though?

lnvoker:
No way around that.
1) If you disconnect (willingly or kicked), plugin won't save your data anywhere.
2) Ship is being automatically removed from carrier's docked ships list after disconnect.
3) With latest commit, switching character leads to same consequences as disconnect.

PR #91, opened 2019-06-24
Docking modules fix

lnvoker:
Main changes:
1) Now players won't be redirected to last real base they were on if they died after being docked to carrier. This part of code suspected to be responsible for character-erasing bug.
2) Conn exploit check removed, jumps to conn were restricted long time ago, so I think it's fine.

### DIFF
--- a/Plugins/Public/mobiledocking_plugin/Main.cpp
+++ b/Plugins/Public/mobiledocking_plugin/Main.cpp
@@ -156,30 +156,12 @@ void __stdcall BaseExit(uint iBaseID, uint iClientID)
 
 }
 
-void __stdcall DisConnect(uint iClientID, enum EFLConnection p2)
-{
-	returncode = DEFAULT_RETURNCODE;
-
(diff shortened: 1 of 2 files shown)


Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -111,37 +111,6 @@ uint GetInstalledModules(uint iClientID)
...
-	FreeWString(flStr);
-
-	// Overwrite the ban file so that it contains the ban reason
-	wstring wscDir;
-	HkGetAccountDirName(acc, wscDir);
-	string scBanPath = scAcctPath + wstos(wscDir) + "\\banned";
-	FILE *file = fopen(scBanPath.c_str(), "wbe");
-	if (file)
-	{
-		fprintf(file, "Autobanned by Docking Module\n");
```
dsyalex: Did a search, no matches. Tested locally and there does not seem to be a way to trigger this behaviour in an abusive way.

PR #100, opened 2019-07-20
playercntl: Pimp-Ship++

lnvoker:
Main changes:
1) No need of kick to set new lights.
2) Able to select few hardpoints.
3) No more resetting. /pimpship commands available once entered specific room.
4) Opportunity to select lights by name, rather than by ID.
5) You can show not all available items, but part of them. More convenience.


Plugins/PluginUtilities/DynPacket.h
```diff
@@ -0,0 +1,270 @@
...
+{
+private:
+	uint Size;
+
+	byte kind;
+	byte type;
+
+public:
+	// This is content of your packet. You may want to reinterpretate it as pointer to packet data struct for convenience.
+	byte content[1];
```
dsyalex: I'm struggling to get my head around the `[1]` here. Shouldn't it just be a pointer?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -180,6 +179,23 @@ namespace PimpShip
 			}
 		}
 
+		mapInfo[iClientID].bInPimpDealer = true;
+
+		// Build the equipment list.
+		int iSlotID = 1;
+
+		list<EquipDesc> &eqLst = Players[iClientID].equipDescList.equip;
+		for (list<EquipDesc>::iterator eq = eqLst.begin(); eq != eqLst.end(); eq++)
```
dsyalex: Before we enter this loop we should probably `mapInfo[iClientID].mapCurrEquip.clear();`?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -192,58 +208,35 @@ namespace PimpShip
...
-			}
+			PrintUserCmdText(iClientID, L"ERR Pimpship facilities are not available here");
+			return true;
 		}
 
-		mapInfo[iClientID].mapCurrEquip.clear();
-		mapInfo[iClientID].bInPimpDealer = true;
-
 		PrintUserCmdText(iClientID, L"Available ship pimping commands:");
+		PrintUserCmdText(iClientID, L"This facility costs " + ToMoneyStr(set_iCost) + L" credits to use per one item.");
```
dsyalex: TODO: did we used to charge per-transaction and it's now per-item? that could be problematic

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -192,58 +208,35 @@ namespace PimpShip
...
 
 		PrintUserCmdText(iClientID, L"/showsetup");
 		PrintUserCmdText(iClientID, L"|     Display current ship setup.");
 
-		PrintUserCmdText(iClientID, L"/showitems");
+		PrintUserCmdText(iClientID, L"/showitems [from]-[to]");
 		PrintUserCmdText(iClientID, L"|     Display items that may be added to your ship.");
```
dsyalex: Document what from and to are.

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -192,58 +208,35 @@ namespace PimpShip
...
 
-		PrintUserCmdText(iClientID, L"/setitem <hardpoint id> <new item id>");
-		PrintUserCmdText(iClientID, L"|     Change the item at <hp id> to <item id>.");
-		PrintUserCmdText(iClientID, L"|     <hi id>s are shown by typing /show setup.");
+		PrintUserCmdText(iClientID, L"/setitem <hardpoint id> <item id>");
+		PrintUserCmdText(iClientID, L"|     Change the item at <hardpoint id> to <item id>.");
+		PrintUserCmdText(iClientID, L"|     <hardpoint id>s are shown by typing /show setup.");
 		PrintUserCmdText(iClientID, L"|     <item id>s are shown by typing /show items.");
+		PrintUserCmdText(iClientID, L"|     Allowed to print item name instead of ID.");
```
dsyalex: Let's use 'item' instead of 'item id' then, and say it can be name or ID shown on /show items?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -192,58 +208,35 @@ namespace PimpShip
...
 
-		PrintUserCmdText(iClientID, L"/buynow");
-		PrintUserCmdText(iClientID, L"|     Confirms the changes.");
-		PrintUserCmdText(iClientID, L"This facility costs " + ToMoneyStr(set_iCost) + L" credits to use.");
+		PrintUserCmdText(iClientID, L"/setitem [hp1]-[hp2] <item id>");
```
dsyalex: Document hp1 and hp2.

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -192,58 +208,35 @@ namespace PimpShip
...
+		PrintUserCmdText(iClientID, L"|     Change items at hardpoints in range of <hp1>-<hp2> to <item id>.");
 
-		wstring wscCharName = (const wchar_t*) Players.GetActiveCharacterName(iClientID);
+		PrintUserCmdText(iClientID, L"/setitem [hp-begin]*<every-n>*[hp-end] <item id>");
+		PrintUserCmdText(iClientID, L"|     Change every n hardpoint to <item id>.");
+		PrintUserCmdText(iClientID, L"|     Changing begins from [hp-begin] if the parameter is determined, otherwise begins from first hardpoint.");
```
dsyalex: if parameter is given?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -271,10 +264,60 @@ namespace PimpShip
 		if (!mapInfo[iClientID].bInPimpDealer || !set_bEnablePimpShip)
 			return false;
```
dsyalex: Note to self: Review from here downwards

Plugins/PluginUtilities/DynPacket.h
lnvoker: We need [1] as flexible array member to make packet struct contiguous in memory.
Pointer would split packet into two allocated memory zones.

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: Oh, yes. And rebuild equip list after purchasing new ship.

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: I was convinced previous version took credits for every execution of /setitem command.

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -271,10 +279,60 @@ namespace PimpShip
...
+			if (*it == '-')
+			{
+				if (it == wscParam.begin() && it == wscParam.end() - 1)
+				{
+					beginFrom = 1;
+					endAt = mapInfo[iClientID].mapCurrEquip.size();
+				}
+				else if (it == wscParam.end() - 1)
+				{
+					beginFrom = ToUInt(wscParam.substr(0, index));
```
dsyalex: There's a lot of these types of calls in this code - what should we do if the user tries to put in non-numeric characters?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -271,10 +279,60 @@ namespace PimpShip
...
+					endAt = mapInfo[iClientID].mapCurrEquip.size();
+				}
+				else if (it == wscParam.begin())
+				{
+					beginFrom = 1;
+					endAt = ToUInt(wscParam.substr(index));
+
+					if (endAt < beginFrom)
+						swap(endAt, beginFrom);
```
dsyalex: where is this defined?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -271,10 +279,60 @@ namespace PimpShip
...
+		}
+
+		if (endAt > mapAvailableItems.size())
+			endAt = mapAvailableItems.size();
+
+		if (beginFrom == 0)
+			beginFrom = 1;
+
+		if (beginFrom > endAt)
+			beginFrom = endAt + 1;
```
dsyalex: this seems wrong? shouldn't we swap beginFrom and endAt in this case?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -271,10 +279,60 @@ namespace PimpShip
...
+
+		PrintUserCmdText(iClientID, L"Showed %u/%u items:", endAt - beginFrom + 1, mapAvailableItems.size());
```
dsyalex:
```suggestion
		PrintUserCmdText(iClientID, L"Showing %u/%u items:", endAt - beginFrom + 1, mapAvailableItems.size());
```

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
+					{
+						beginFrom = ToUInt(firstArg.substr(0, index));
+						everyN = ToUInt(firstArg.substr(index));
+
+						if (beginFrom == 0)
+							break;
+					}
+				}
+				else
+				{
```
dsyalex: we don't set everyN in this case?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
-			i != mapInfo[iClientID].mapCurrEquip.end(); ++i)
+		if (beginFrom == 0 || beginFrom > mapInfo[iClientID].mapCurrEquip.size())
 		{
-			pub::Player::RemoveCargo(iClientID, i->second.sID, 1);
+			PrintUserCmdText(iClientID, L"ERR Beginning is out of bounds");
+			PrintUserCmdText(iClientID, L"You may want to use following syntax to select all hardpoints from beginning to %u:", endAt);
+			if (everyN == 1)
+				PrintUserCmdText(iClientID, L"/setitem -%u %u", endAt, iSelectedItemID);
+			else
+				PrintUserCmdText(iClientID, L"/setitem *%u*%u %u", everyN, endAt, iSelectedItemID);
```
dsyalex:
```suggestion
				PrintUserCmdText(iClientID, L"/setitem -%u*%u %u", endAt, everyN, iSelectedItemID);
```
?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
-			i != mapInfo[iClientID].mapCurrEquip.end(); ++i)
+		if (endAt > mapInfo[iClientID].mapCurrEquip.size())
 		{
-			HkAddEquip(wscCharName, i->second.iArchID, wstos(i->second.wscHardPoint));
+			PrintUserCmdText(iClientID, L"ERR Ending is out of bounds");
+			PrintUserCmdText(iClientID, L"You may want to use following syntax to select all hardpoints from %u to end:", beginFrom);
+			if (everyN == 1)
+				PrintUserCmdText(iClientID, L"/setitem %u- %u", beginFrom, iSelectedItemID);
+			else
+				PrintUserCmdText(iClientID, L"/setitem %u*%u* %u", beginFrom, everyN, iSelectedItemID);
```
dsyalex:
```suggestion
				PrintUserCmdText(iClientID, L"/setitem %u-*%u %u", beginFrom, everyN, iSelectedItemID);
```

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
+		int totalCost = 0;
+		map<uint, EQ_HARDPOINT>& info = mapInfo[iClientID].mapCurrEquip;
+		uint newItem = mapAvailableItems[iSelectedItemID].iArchID;
+		for (uint i = beginFrom; i < endAt + 1; i += everyN)
+		{
+			if (info[i].iArchID != newItem)
+				totalCost += set_iCost;
+		}
+
+		if (totalCost == 0)
```
dsyalex: This will break if the cost is configured as 0.

Plugins/Public/playercntl_plugin/PimpShip.cpp
dsyalex: No, I am wrong. We should talk about whether the syntax is too complicated though.

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -271,10 +279,60 @@ namespace PimpShip
...
+		uint endAt = mapAvailableItems.size();
+
+		ushort index = 0;
+		for (wstring::const_iterator it = wscParam.begin(); it != wscParam.end(); it++)
+		{
+			index++;
+
+			if (*it == '-')
+			{
+				if (it == wscParam.begin() && it == wscParam.end() - 1)
```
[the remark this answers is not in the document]
lnvoker: ...yes. Should I change that?

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: It will be parsed as 0, doesn't hurt actually.

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: swap belongs to std namespace

Plugins/Public/playercntl_plugin/PimpShip.cpp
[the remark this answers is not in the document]
dsyalex: I think it's fine, just checking my understanding

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: No. All swapping is done before. It checks if beginning is not getting out of list bounds (more than 137).

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: No. if everyN is set, that means beginFrom is already set to be more than 0 too.

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: Going to fix.

lnvoker: Fixed.

Plugins/Public/playercntl_plugin/PimpShip.cpp
dsyalex: So this could be `if (beginFrom > endAt + 1)`, but ok

Plugins/Public/playercntl_plugin/PimpShip.cpp
dsyalex: Decided we're going to divide the price by 10 in the config.

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
+				if (beginFrom == 0)
+				{
+					endAt = mapInfo[iClientID].mapCurrEquip.size();
+
+					if (it == firstArg.begin())
+					{
+						beginFrom = 1;
+						everyN = ToUInt(firstArg.substr(index));
+					}
+					else if (it == firstArg.end() - 1)
```
dsyalex: in the event that this is the first * but it's also the end, shouldn't we just error out right there?

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
 		}
 
-		// Remove all lights.
-		for (map<uint, EQ_HARDPOINT>::iterator i = mapInfo[iClientID].mapCurrEquip.begin();
-			i != mapInfo[iClientID].mapCurrEquip.end(); ++i)
+		if (beginFrom == 0 || beginFrom > mapInfo[iClientID].mapCurrEquip.size())
 		{
-			pub::Player::RemoveCargo(iClientID, i->second.sID, 1);
+			PrintUserCmdText(iClientID, L"ERR Beginning is out of bounds");
+			PrintUserCmdText(iClientID, L"You may want to use following syntax to select all hardpoints from beginning to %u:", endAt);
```
dsyalex: only `all` if everyN is 1

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,220 @@ namespace PimpShip
...
 		}
 
-		// Re-add all lights so that the order is kept the same
-		for (map<uint, EQ_HARDPOINT>::iterator i = mapInfo[iClientID].mapCurrEquip.begin();
-			i != mapInfo[iClientID].mapCurrEquip.end(); ++i)
+		if (endAt > mapInfo[iClientID].mapCurrEquip.size())
 		{
-			HkAddEquip(wscCharName, i->second.iArchID, wstos(i->second.wscHardPoint));
+			PrintUserCmdText(iClientID, L"ERR Ending is out of bounds");
+			PrintUserCmdText(iClientID, L"You may want to use following syntax to select all hardpoints from %u to end:", beginFrom);
```
dsyalex: only `all` if everyN is 1

dsyalex:
I think I've got my head around the syntax stuff now.

Plugins/Public/playercntl_plugin/PimpShip.cpp
lnvoker: The syntax allows to execute the command without 2 arguments, only everyN is non-optional.
*2* = 2* = *2 = 1*2 = 1*2* = *2*10 = 1*2*10
A lot of ways you see.

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,230 @@ namespace PimpShip
...
-			pub::Player::RemoveCargo(iClientID, i->second.sID, 1);
+			PrintUserCmdText(iClientID, L"ERR Beginning is out of bounds");
+			if (everyN == 1)
+			{
+				PrintUserCmdText(iClientID, L"You may want to use following syntax to select all hardpoints from beginning to %u:", endAt);
+				PrintUserCmdText(iClientID, L"/setitem -%u %u", endAt, iSelectedItemID);
+			}
+			else
+			{
+				PrintUserCmdText(iClientID, L"You may want to use following syntax to select hardpoints from beginning to %u:", endAt);
```
dsyalex:
```suggestion
				PrintUserCmdText(iClientID, L"You may want to use following syntax to select every %u hardpoints from beginning to %u:", everyN, endAt);
```

Plugins/Public/playercntl_plugin/PimpShip.cpp
```diff
@@ -287,69 +345,230 @@ namespace PimpShip
...
-			HkAddEquip(wscCharName, i->second.iArchID, wstos(i->second.wscHardPoint));
+			PrintUserCmdText(iClientID, L"ERR Ending is out of bounds");
+			if (everyN == 1)
+			{
+				PrintUserCmdText(iClientID, L"You may want to use following syntax to select all hardpoints from %u to end:", beginFrom);
+				PrintUserCmdText(iClientID, L"/setitem %u- %u", beginFrom, iSelectedItemID);
+			}
+			else
+			{
+				PrintUserCmdText(iClientID, L"You may want to use following syntax to select hardpoints from %u to end:", beginFrom);
```
dsyalex:
```suggestion
				PrintUserCmdText(iClientID, L"You may want to use following syntax to select every %u hardpoints from %u to end:", everyN, beginFrom);
```

PR #98, opened 2019-07-16
npc_control: Advanced NPC disabling

lnvoker:
Main changes:
1) Allows specific NPCs to spawn even at high server load.
2) Replaces basic FLHook NPC disabling. Mission and .aicreate NPCs still spawn at any server load.

PR #103, opened 2019-08-20
Revert "npc_control: Advanced NPC disabling"

dsyalex:
Reverts DiscoveryGC/FLHook#98

We ran into loads of problems with this, there seems to be all sorts of hidden compatibility and memory usage issues associated with it.


lnvoker:
There seems to be no way to make my NPC disabling work properly.
Thus, after dozen tests it was decided to remove the thing and revert npc_control to its previous state.

To have spawning NPCs at high server load is still possible, there exist ways around the problem.

PR #104, opened 2019-08-26
npc-control: Added optional parameter to .aifollow to specify which character to follow

opechey:
**Examples:**

_.aifollow_ (Will follow the admin and send a message to command giver saying "Following Player1".
_.aifollow player2_ (Will follow the 2nd player and send a message to command giver  saying "Following Player2".

If you try to use this on a character that is not online you get: "Player2 is not online".

If you try and use this on a character that is not in space you get: "Player2 is not in space"

### DIFF
--- a/Plugins/Public/npc_control/Main.cpp
+++ b/Plugins/Public/npc_control/Main.cpp
@@ -631,7 +631,7 @@ void AdminCmd_AICome(CCmds* cmds)
 }
 
 /* Make AI follow you until death */
-void AdminCmd_AIFollow(CCmds* cmds)
+void AdminCmd_AIFollow(CCmds* cmds, wstring Charname)
 {
 	if (!(cmds->rights & RIGHT_SUPERADMIN))
 	{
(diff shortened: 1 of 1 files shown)


Plugins/Public/npc_control/Main.cpp
```diff
@@ -828,7 +845,7 @@ bool ExecuteCommandString_Callback(CCmds* cmds, const wstring &wscCmd)
 	else if (IS_CMD("aifollow"))
 	{
 		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
-		AdminCmd_AIFollow(cmds);
+		AdminCmd_AIFollow(cmds, cmds->ArgStr(1));
```
dsyalex: what about ArgCharname? that would mean you'd do `.aifollow!` to get yourself instead of just `.aifollow` but it reuses all the existing code

opechey: Thanks @dsyalex, I have swapped this over and confirmed it works. Would you prefer I still keep the check for no parameter and set to admin so `.aifollow` works as well as `.aifollow!` or strip this out?

PR #105, opened 2019-08-26
npc-control: Added 2 optional infocard parameters

opechey:
Added 2 optional infocard parameters to npc in alley_npc.cfg. Either one or two infocard parameters may be used. If none are specified it will generate random names as per previous functionality.

### DIFF
--- a/Plugins/Public/npc_control/Main.cpp
+++ b/Plugins/Public/npc_control/Main.cpp
@@ -73,6 +73,8 @@ struct NPC_ARCHTYPESSTRUCT
 	uint Shiparch;
 	uint Loadout;
 	uint IFF;
+	uint Infocard;
+	uint Infocard2;
 	int Graph;
 };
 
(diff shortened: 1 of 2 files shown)


Plugins/Public/npc_control/Main.cpp
```diff
@@ -503,8 +509,16 @@ void CreateNPC(wstring name, Vector pos, Matrix rot, uint iSystem)
 		// below shows the use of multiple part names.
 		FmtStr pilot_name(0, 0);
 		pilot_name.begin_mad_lib(16163); // ids of "%s0 %s1"
-		pilot_name.append_string(rand_name());  // ids that replaces %s0
-		pilot_name.append_string(rand_name()); // ids that replaces %s1
+		if (arch.Infocard != 0) {
```
dsyalex: hm, can we rely on uint in a struct being initialised to 0? this feels like potentially UB

Plugins/Public/npc_control/Main.cpp
```diff
@@ -503,8 +509,16 @@ void CreateNPC(wstring name, Vector pos, Matrix rot, uint iSystem)
...
+			pilot_name.append_string(arch.Infocard);
+			if (arch.Infocard2 != 0) {
```
dsyalex: ^

Plugins/Public/npc_control/Main.cpp
opechey: Hi @dsyalex, I have added initialization of the ints just before they get set. Hope this is okay.

PR #106, opened 2019-09-14
Balance Magic cleanup

lnvoker:
1) Cleanup:
 a. Incomplete Burstfire removed. Why do we even have basically not working incomplete feature in release version? Read notes.
 b. This socket thing removed.
 c. Debug messages removed.
 d. Other redundant parts of code or comments removed.
2) Optimization:
 a. Damage is being changed by multipliers rather than by adjust values.
 b. No need to load list of armor items and scan damaged ships for them every landed shot.
3) Bugfix:
 a. Wrong shield rebuild time - fixed.
 b. Wrong death message - fixed.
4) New features:
 a. Damage dealt to solar objects is being changeable separately.

Notes:
- If you wish to get removed code - look in git history.
- Someone has to make changes in config file for plugin to work properly.


Plugins/Public/balancemagic/Main.cpp
```diff
@@ -345,296 +198,69 @@ bool UserCmd_Process(uint iClientID, const wstring &wscCmd)
...
-	if (obj)
-	{
-		CShip* cship = (CShip*)HkGetEqObjFromObjRW((IObjRW*)obj);
-
-		for (short* slot = wpn.sArray1; slot != wpn.sArray2; slot++)
-		{
-			CEquipManager* eqmanager = (CEquipManager*)((char*)cship + 0xE4);
-			CEquip *equip = eqmanager->FindByID(*slot);
-			if (CEGun::cast(equip))
+			// Deduce: if not fighter nor freighter, then it's obviously solar object.
```
dsyalex: this sounds dangerous - there are lots of other types of space objects defined in FLCoreCommon.h

Plugins/Public/balancemagic/Main.cpp
```diff
@@ -345,296 +198,69 @@ bool UserCmd_Process(uint iClientID, const wstring &wscCmd)
...
-					aClientData[iClientID].mapBurstFills[i].count = iter->second.count;
-					aClientData[iClientID].mapBurstFills[i].timestamp = 0;
-					i++;
-				}
+				uint iArchID;
+				pub::SpaceObj::GetSolarArchetypeID(iDmgToSpaceID, iArchID);
+				uint targetShipClass = Archetype::GetShip(iArchID)->iShipClass;
+
+				if (targetShipClass == 0 || targetShipClass == 1 || targetShipClass == 3)
+					setHealth = curr - (curr - setHealth) * iter->second.fighter;
```
dsyalex: don't we still need the separate case for shields being up?

Plugins/Public/balancemagic/Main.cpp
dsyalex: We discussed this and it's fine. FL only lets us have fighters, freighters, and solars. And probably loot in space but we don't care about that.

Plugins/Public/balancemagic/Main.cpp
dsyalex: Discussed this and FL actually handles this all for us, it seems fine.

PR #107, opened 2019-09-21
HyperJump (1/7)

lnvoker:
1. Removed:
 a. Commented out code.
 b. Death and banned systems.
 c. Detection of macro users (code to catch assholes) as it may ban users who just use Ctrl + ↑ to quickly pull coordinates from chat history.
 d. .testjump command
2. Added:
 a. System whitelist checks at blind jumps and /setcoords command.
 b. Inability to use /setcoords while jumping.

### DIFF
--- a/Plugins/Public/playercntl_plugin/HyperJump.cpp
+++ b/Plugins/Public/playercntl_plugin/HyperJump.cpp
@@ -36,7 +36,7 @@ namespace HyperJump
 			ConPrint(L"ERROR: item '%s' is not valid\n", stows(nickname).c_str());
 		}
 
-		return item;
+		return item;	
 	}
 
 	// Ships restricted from jumping
@@ -171,9 +171,9 @@ namespace HyperJump
 	{
 		wstring sbuf;
 		wchar_t buf[100];
-		for (int i = 0; i < HCOORD_SIZE; i++)
+		for (int i=0; i<HCOORD_SIZE; i++)
 		{
-			if (i != 0 && (i % 4) == 0) sbuf += L"-";
+			if (i!=0 && (i%4)==0) sbuf += L"-";
 			_snwprintf(buf, sizeof(buf), L"%02X", (byte)ibuf[i]);
 			sbuf += buf;
 		}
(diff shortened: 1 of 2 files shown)


Plugins/Public/playercntl_plugin/HyperJump.cpp
```diff
@@ -1903,25 +1580,39 @@ namespace HyperJump
...
-			while(isBannedSystem)
+			if (!JumpSystemListEnabled)
 			{
+				vector<string> systems;
+				struct Universe::ISystem *sysinfo = Universe::GetFirstSystem();
+				while (sysinfo)
+				{
+					systems.push_back(sysinfo->nickname);
+					sysinfo = Universe::GetNextSystem();
+				}
```
dsyalex: This block could be:

```suggestion
				for (struct Universe::ISystem *sysinfo = Universe::GetFirstSystem(); sysinfo; sysinfo = Universe::GetNextSystem())
				{
					systems.push_back(sysinfo->nickname);
				}
```

Plugins/Public/playercntl_plugin/HyperJump.cpp
```diff
@@ -1798,56 +1507,25 @@ namespace HyperJump
...
 			}
-		////////////////////////// System limit restriction ///////////////////////////////////////
-		if (JumpSystemListEnabled == 1 && jd.iTargetSystem != 0)
-		{
-			const Universe::ISystem *iSys = Universe::get_system(iSystemID);
-		    wstring wscSysName = HkGetWStringFromIDS(iSys->strid_name);
-			//PrintUserCmdText(iClientID, L"Space kitteh knows you are in %u !", iSystemID);
-			for (map<uint, SYSTEM_JUMPLIST>::iterator iter = mapJumpSystems.begin(); iter!=mapJumpSystems.end(); iter++)
+			////////////////////////// System limit restriction ///////////////////////////////////////
+			else if (JumpSystemListEnabled == 1 && jd.iTargetSystem != 0)
```
dsyalex: `jd.iTargetSystem != 0` is redundant in this case, this is an `else if` where the original `if` condition was `jd.iTargetSystem == 0`

dsyalex: Also, checking a bool is equal to 1 seems strange, just directly use the boolean

Plugins/Public/playercntl_plugin/HyperJump.cpp
```diff
@@ -1727,13 +1423,27 @@ namespace HyperJump
 			PrintUserCmdText(iClientID, L"ERR Invalid coordinates, parity error");
 			return true;
 		}
-	
+
+		////////////////////////// System limit restriction ///////////////////////////////////////
+		if (JumpSystemListEnabled == 1)
```
dsyalex: Checking a bool is equal to 1 seems strange, just directly use the boolean.

dsyalex:
Otherwise looks good.

PR #113, opened 2020-04-15
FLHook, base: Add POB module query IPC, fabrication module

Kazinsal:
The Equipment Fabrication Bay is a module that does nothing, but will be used in conjunction with the newly-added CUSTOM_BASE_QUERY_MODULE inter-plugin communication to implement equipment fabrication on POBs as well as NPC stations.

**This change requires an FLHook rebuild due to the addition of a new inter-plugin communication call.**


dsyalex (changes requested):
Needs rebase

dsyalex:
This is now hopelessly outdated

PR #115, opened 2020-04-24
pvecontroller: Add PvE Overhaul Phase 1 code.

Kazinsal:
Take three at actually getting my branches to not break horribly...

This PR is for the initial plugin work for the PvE controller. It is feature complete up to the bits required for the Rheinland Update (and as of this commit, for using system bonus multipliers for ie. events). This plugin is new but has been fairly extensively tested on the development test server in many different scenarios and group sizes.


Kazinsal:
Reviewed with @dsyalex, commit cff45d5 resolves issues discovered and refactors a few bits of code to be less reliant on magic numbers and hardcoded FLHashIDs. Pending review and submission.

PR #117, opened 2020-06-29
PvE Controller Warzones

Corran-Raisu:
Currently. there are only system-wide multipliers which globally effect bounty rewards for everyone regardless of faction.

With the added war zone spawns for NPCs, some players take advantage of this by using ships hostile to both factions in order to farm the higher spawn rates. 

This allows affiliation specific multipliers which can be used alongside system multipliers to balance / nerf / buff bounty rewards between factions involved in a warzone vs factions not involved in a warzone.

### DIFF
--- a/Plugins/Public/pvecontroller/Main.cpp
+++ b/Plugins/Public/pvecontroller/Main.cpp
@@ -43,12 +43,19 @@ struct stDropInfo {
 	float fChance;
 };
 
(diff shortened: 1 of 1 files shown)


dsyalex (dismissed):
Needs rebase

Kazinsal (approved):
Looks good. This will be handy for funneling activity.

PR #118, opened 2020-09-03
Class Difference Multipliers

Corran-Raisu:
Bounty Payout will be multiplied depending on the difference between ship class that killed the target NPC and the ship class of the target NPC.

Important:  LowerBound is used, while the code doesn't apply a modifier of 1.0, an entry that has a 1.0 modifier is needed.

Example Config entries:

; This groups the "shipclass" used in shiparchs into categories, or types, to simplify comparison for class-diff and allow the groupings themselves to be adjusted.
; class_type = class, ship_class1, ship_class2, ship_classN (can have as many or as few
class_type = 0, 0, 1, 3, 19
class_type = 1, 2, 4, 5, 6, 7, 8, 9, 10
class_type = 2, 11, 12
class_type = 3, 13, 14
class_type = 4, 15
class_type = 5, 16, 17, 18

; Scaling settings for payouts based on class difference of player that killed NPC.
; class_diff = diff, multiplier
class_diff = -2, 0.5
class_diff = -1, 0.8
class_diff = 0, 1
class_diff = 1, 1.12
class_diff = 2, 1.25
class_diff = 3, 1.5

### DIFF
--- a/Plugins/Public/pvecontroller/Main.cpp
+++ b/Plugins/Public/pvecontroller/Main.cpp
@@ -52,6 +52,8 @@ map<uint, float> mapBountySystemScales;
 list<uint> lstRecordedBountyObjs;
 
 multimap<uint, stDropInfo> mmapDropInfo;
+map<uint, uint> mapShipClassTypes;
+map<int, float> mapClassDiffMultipliers;
 
 int set_iPluginDebug = 0;
 float set_fMaximumRewardRep = 0.0f;
@@ -64,6 +66,8 @@ int iLoadedNPCShipBountyOverrides = 0;
 int iLoadedNPCBountyGroupScale = 0;
 int iLoadedNPCBountyArmorScales = 0;
 int iLoadedNPCBountySystemScales = 0;
+int iLoadedClassTypes = 0;
+int iLoadedClassDiffMultipliers = 0;
 void LoadSettingsNPCBounties(void);
 
 
@@ -115,6 +119,10 @@ void LoadSettingsNPCBounties()
 	iLoadedNPCBountyArmorScales = 0;
 	mapBountySystemScales.clear();
 	iLoadedNPCBountySystemScales = 0;
+	mapShipClassTypes.clear();
+	iLoadedClassTypes = 0;
+	mapClassDiffMultipliers.clear();
+	iLoadedClassDiffMultipliers = 0;
 
 	// Load ratting bounty settings
 	set_iPoolPayoutTimer = IniGetI(scPluginCfgFile, "NPCBounties", "pool_payout_timer", 0);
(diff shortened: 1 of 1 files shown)


dsyalex (dismissed):
Your description does not talk about the new `class_type` field

Plugins/Public/pvecontroller/Main.cpp
```diff
@@ -572,6 +600,26 @@ void __stdcall HkCb_AddDmgEntry(DamageList *dmg, unsigned short p1, float damage
...
+
+					int classDiff = 0;
+					map<uint, uint>::iterator itVictimType = mapShipClassTypes.find(victimShiparch->iShipClass);
+					map<uint, uint>::iterator itKillerType = mapShipClassTypes.find(iDmgFromShipClass);
+					if (itVictimType != mapShipClassTypes.end() && itKillerType != mapShipClassTypes.end())
+						classDiff = itVictimType->second - itKillerType->second;
+
+					map<int, float>::iterator itDiffMultiplier = mapClassDiffMultipliers.lower_bound(classDiff);
+					if (itDiffMultiplier != mapClassDiffMultipliers.end())
+						if (itDiffMultiplier->second != 1.0f) //We only need to modify the payout if the multiplier is not 1.
```
dsyalex, line 615: to avoid a multiplication by 1...? think we can skip this check

Corran-Raisu: Removed in latest commit.

Corran-Raisu:
> Your description does not talk about the new `class_type` field

Forgot to reply on this specifically: I've added info about class_type in description. 

If it sees potential usage outside of PVEController, this Ship Class to Ship Type conversion would probably be better put somewhere else in FL Hook for more general access rather than it being in a specific plugin.

Plugins/Public/pvecontroller/Main.cpp
```diff
@@ -572,6 +600,23 @@ void __stdcall HkCb_AddDmgEntry(DamageList *dmg, unsigned short p1, float damage
...
+					map<uint, uint>::iterator itVictimType = mapShipClassTypes.find(victimShiparch->iShipClass);
+					map<uint, uint>::iterator itKillerType = mapShipClassTypes.find(iDmgFromShipClass);
+					if (itVictimType != mapShipClassTypes.end() && itKillerType != mapShipClassTypes.end())
+						classDiff = itVictimType->second - itKillerType->second;
+
+					map<int, float>::iterator itDiffMultiplier = mapClassDiffMultipliers.lower_bound(classDiff);
+					if (itDiffMultiplier != mapClassDiffMultipliers.end())
+						iBountyPayout *= itDiffMultiplier->second;
+						if (set_iPluginDebug >= PLUGIN_DEBUG_VERYVERBOSE)
+							PrintUserCmdText(iDmgFrom, L"PVECONTROLLER: Modifying payout to $%d (%0.2f of normal) due to class difference. %u vs %u \n", iBountyPayout, itDiffMultiplier->second, itKillerType->second, itVictimType->second);
```
dsyalex, line 617:
```suggestion
							PrintUserCmdText(iDmgFrom, L"PVECONTROLLER: Modifying payout to $%d (%0.2f x normal) due to class difference. %u vs %u \n", iBountyPayout, itDiffMultiplier->second, itKillerType->second, itVictimType->second);
```

Kazinsal:
Looks good to me. This might help a bit PvE balancing a bit, and give incentive to not just throw battleships at everything.

dsyalex (approved):

PR #119, opened 2020-09-03
Base will allow to buy stuff regardless to min_stock restriction

dd84ai:
Suggesting so the player logged as admin could buy stuff regardless of min_stock restrictions. I changed shop settings too much, just to buy stuff and lock it back as unbuyable.


dsyalex:
This change looks good, there's an issue with the existing code which we're discussing on Discord

dsyalex (approved):

dsyalex:
let's get this merged first

PR #124, opened 2020-09-10
'set chattime on' now now not ignoring 'diemsg'-setting anymore

PhoenixxClaw:
fix for https://discoverygc.com/forums/showthread.php?tid=182469

Deathtime now won't be posted unless '/set dietime' is enabled. (as intended)


dsyalex (changes requested):
Per discussion on Discord, this will break `!set_bUserCmdSetDieMsg` behaviour, what was actually wrong there was the missing bShowChatDieTime check etc.
Instead you should move the DIEMSG_ALL check up to the top of the if blocks (i.e. before DIEMSG_NONE), and add a disjunction against `!set_bUserCmdSetDieMsg`

dsyalex (approved):

PR #125, opened 2020-09-12
Hyper scanner fix

dd84ai:
You can /survey while being bumped by NPC within 400 distance by x, y, z (not 5 as before)

### DIFF
--- a/Plugins/Public/playercntl_plugin/HyperJump.cpp
+++ b/Plugins/Public/playercntl_plugin/HyperJump.cpp
@@ -638,7 +638,7 @@ namespace HyperJump
 				SURVEY &sm = iter->second;
 				if (sm.charging_on)
 				{
-					//remember starting location
+					//Remember starting location of survey
 					Matrix ornt;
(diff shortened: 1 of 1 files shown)


dd84ai:
This fix would allow to be not disrupting hyper space survey when you are bumped by different NPCs around during hyper

dsyalex (changes requested):
Per Discord, this code looks like it will allow charging up to 400m/s (i.e., it's fine to charge while cruising :D), as it's actually checking speed of the ship. What you want is distance from the charge start location :)

dd84ai:
changed

dsyalex (changes requested):
while curr_charge <= charge_rate you'll get pos_start == pos_curr, but as soon as curr_charge > charge_rate you'll get pos_start = {0, 0, 0} at which point you won't be able to charge jump drive except inside the sun

dd84ai:
fixed start_pos

dsyalex (changes requested):
HkDistance3D please

dsyalex (approved):

PR #126, opened 2020-09-14
Jumpbeacon fix

PhoenixxClaw:


PhoenixxClaw:
As explained in our Discord Convo.
The Vagrant JD works just fine. There was a nullpntr that happened when trying to jump out of (or into) a restricted system via the '/jumpbeacon PLAYERNAME' command, thanks to 'jd.iTargetSystem' only getting set if you jump via '/setcords'.

dsyalex (changes requested):
I think you should either remove (new) line 1874, or remove the line you've added and change (new) 1861 to drop `jd.`

PhoenixxClaw:
tbh, I didn't notice that there is a second line, stating exactly what I did at 1853. Adjusted that, PR updated

dsyalex (approved):

dsyalex:
needs a rebase. please also clean up the history. you could start from a clean copy of the master branch and force-push over this

Issue #47, opened 2018-04-29
POBs sometimes complain about reps when they shouldn't


dsyalex:
When docking
Happens only sometimes, but is a right pain in the ass when it does happen.

dd84ai:
Resolved issue: https://github.com/DiscoveryGC/FLHook/pull/128

dsyalex:
Fixed by #128

PR #128, opened 2020-09-16
base: fix for docking bug, a.k.a. "Your reputation stinks" (fixed 12.10.20)

dd84ai:
Fix to pob docking problem https://github.com/DiscoveryGC/FLHook/issues/47
It prevents playercntl plugin interference, which incorrect worked caused "your reputation stinks" and etc during pob dock randomly


dd84ai:
wait a second, it has not solved problems yet

dd84ai:
now it is working regardless of which plugin is loaded first. done,

dsyalex (changes requested):
please squash your commits and add a reasonable description

dd84ai:
Resolved

dd84ai:
Discovered it can be coded possibly easier than I did

dd84ai:
Remade removing not optimized trash ;b

Source/FLHookPluginSDK/headers/plugin.h
```diff
@@ -328,6 +329,12 @@ struct CUSTOM_BASE_IS_DOCKED_STRUCT
 	uint iDockedBaseID;
 };
 
+struct CUSTOM_BASE_IS_IT_POB_STRUCT
+{
+	uint iBase;
+	bool iAnswer;
```
dsyalex, line 335:
```suggestion
	bool bAnswer;
```

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2497,6 +2497,17 @@ void Plugin_Communication_CallBack(PLUGIN_MESSAGE msg, void* data)
...
 		}
 	}
+	if (msg == CUSTOM_IS_IT_POB)
+	{
+		CUSTOM_BASE_IS_IT_POB_STRUCT* info = reinterpret_cast<CUSTOM_BASE_IS_IT_POB_STRUCT*>(data);
+		PlayerBase *base = GetPlayerBase(info->iBase);
+		returncode = SKIPPLUGINS;
+		if (base)
+		{
+			returncode = SKIPPLUGINS;
```
dsyalex, line 2507:
```suggestion
```

dsyalex (commented):
What about the -0.55 check on POBs? Players shouldn't be able to dock if their reputation with the base's affiliation is too close to hostile right?

dd84ai:
> What about the -0.55 check on POBs? Players shouldn't be able to dock if their reputation with the base's affiliation is too close to hostile right?

This restriction is supposed to be applied only during defensemode == 2, when base allows docking rights from player reputation
Perhaps sometime later, someone could fix it. It is not a really important feature anyway, too overspecialized.
There would be needed changes in IsDockingAllowed function

Source/FLHookPluginSDK/headers/plugin.h
dd84ai: applied

Plugins/Public/base_plugin/Main.cpp
dd84ai: applied

dsyalex (approved):

PR #129, opened 2020-09-16
base: fix for shop bug with minstock (fixed 12.10.20)

dd84ai:
We discussed it before, this is what you suggested Alex.

Better description: there is bug that anyone can buy goods more than minstock allows.
By taking into account how much 'customer' requested, the amount of goods will be never lower than minstock amount.

### DIFF
--- a/Plugins/Public/base_plugin/Main.cpp
+++ b/Plugins/Public/base_plugin/Main.cpp
@@ -1653,7 +1653,7 @@ void __stdcall GFGoodBuy(struct SGFGoodBuyInfo const &gbi, unsigned int client)
 		const wstring &charname = (const wchar_t*)Players.GetActiveCharacterName(client);
 
 		// In theory, these should never be called.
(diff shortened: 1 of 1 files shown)


dsyalex (changes requested):
let's add the brackets to be clear what should be happening here, also fix the description to describe the change please

dd84ai:
Added description and brackets

dsyalex:
`base->market_items[gbi.iGoodID].min_stock > base->market_items[gbi.iGoodID].quantity - count`
there is still ambiguity here, both `>` and `-` in use without brackets

dd84ai:
I get it. in case if the result of > operation can be casted into int and becoming 0 - 1 value used for subtracting operation.

dd84ai:
fixed

dsyalex (commented):
Looks good but we now need to also take care of the check on line 1671

dd84ai:
> Looks good but we now need to also take care of the check on line 1671

fixed

dsyalex (approved):

PR #131, opened 2020-09-21
base: Priority to active recipe over repair (Requires fixing)

dd84ai:
When player builds something, it often requires some precise numbers like... 10'000 Robotic hardware being consumed.
During its crafting he brings exactly 10'000, but recipe consumes only 9'800
because something like 128 units of robotic hardware become consumed for base repair.

This fix prevents it. If the repair materials is still needed in active crafting recipe, it will be not consumed for repairs. Only for crafting. 

Crafting can be still not finished, but if for example need in robotic hardware need is already zero, it will continue being used for repairs.

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -445,6 +445,36 @@ void CoreModule::RepairDamage(float max_base_health)
 
 				if (base->HasMarketItem(item->good) >= item->quantity)
 				{
+					
+					//This cycle prevents base consuming materials for repair if it is needed in active recipe
+					bool found_the_same_consumed_material = false;
+					for (auto i = base->modules.begin();
+						i != base->modules.end(); ++i)
+						if (!(!(*i) ||
+							((*i)->type != Module::TYPE_M_CLOAK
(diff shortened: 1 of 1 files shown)


dsyalex (commented):
This sounds dangerous - while a recipe requiring some repair-commodity is in the process of being made, you can't repair the base?

dd84ai:
It is already dangerous so, because recipe consume commodities 100 times faster than repairs done.
This danger is resolved though in... allowing pause/resume to recipes: https://github.com/DiscoveryGC/FLHook/pull/133

dd84ai:
Anyway, once recipe consumed repair materials, base continues to repair even if recipe is not finished yet.

dsyalex:
So, in other words, repair takes priority at the moment, but at such a slow rate that if an active recipe needs that cargo, the majority of it will go to the recipe before the next repair attempt, not providing a great deal of repair value but annoying the players trying to give it resources for the recipe?

dd84ai:
> So, in other words, repair takes priority at the moment, but at such a slow rate that if an active recipe needs that cargo, the majority of it will go to the recipe before the next repair attempt, not providing a great deal of repair value but annoying the players trying to give it resources for the recipe?

Exactly.

dd84ai:
I realized the lack of compatibility with the pause/resume features.

PR #132, opened 2020-09-21
base: Allow /bank deposit and status without admin access

dd84ai:
Players want the ability to donate to pob banks without admin access.
So if admin is not present, they can donate some money and keep the base floating with supplies it would not accept otherwise.


dd84ai:
I saw no point to forbid /base status command because every player could see money anyway through Esc without admin access already.

Corran-Raisu:
Probably not the best title here.  I see in the commit that the admin check was added to the withdraw command. However the title "/bank withdraw and status without admin access" -- withdraw without admin access would be a huge no.

However, this just allows deposit without admin access, but not withdraw.

dd84ai:
Oh yes, I made a bad choice in naming it properly. fixed.

dsyalex (approved):

PR #134, opened 2020-09-25
Allow TradeEvent to be TargetID of CombatEvents

Corran-Raisu:
This allows the ID of a TradeEvent to be used as a TargetID of a CombatEvent.

### DIFF
--- a/Plugins/Public/event/Main.cpp
+++ b/Plugins/Public/event/Main.cpp
@@ -1328,9 +1328,9 @@ void SendDeathMsg(const wstring &wscMsg, uint iSystem, uint iClientIDVictim, uin
 	string sIDVictimEvent;
 	if (victim)
 	{
-		sIDVictimEvent = wstos(HookExt::IniGetWS(iClientIDVictim, "event.eventid"));
 		if (HookExt::IniGetB(iClientIDVictim, "event.enabled"))
 		{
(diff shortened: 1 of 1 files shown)


dsyalex (changes requested):
shouldn't we be checking event.enabled for this still?
why are we re-using the target ID field instead of making a target event field?

Corran-Raisu:
There wasn't a removal of a check on event.enabled so I'm not sure what you're referring to there.

Reusing TargetID appeared fitting in its usage of both forms of internal nicknames that are used as IDs. 
In addition, this should potentially allow for a CombatEvent where the only valid targets are specifically ships participating in a TradeEvent and possibly avoid confusion in how the two interact together or how they're handled.

It can be changed if necessary though.

dsyalex:
You moved the `wstos(HookExt::IniGetWS(iClientIDVictim, "event.eventid"))` line outside the `if (HookExt::IniGetB(iClientIDVictim, "event.enabled"))` block. Don't we only want this behaviour to trigger if the victim has the event enabled?

Corran-Raisu:
Ah, that. 

Yes, I should have only moved the variable declaration out of the if block so that it could be used later in the Allowed ID vs Target ID portion of the function.

dsyalex (approved):

PR #138, opened 2020-09-28
playercntl: Command /resetrep - to reset all reputations (12.10.20)

dd84ai:
Descritiption: when /resetrep is used. It will nullify all reputations to zero like in fresh restart
What is it for? to restore broken reputations after changing IDs.

Restrictions for usage: 
1) It can be used only in base + it kicks (for logging to rephack your reps according to your ID)
2) It takes some credits (default price 10'000'000)
3) It can be used only once in two weeks, cooldown is resistant to renameme and movechar operations.

### DIFF
--- a/Plugins/Public/playercntl_plugin/MiscCmds.cpp
+++ b/Plugins/Public/playercntl_plugin/MiscCmds.cpp
@@ -64,6 +64,13 @@ namespace MiscCmds
 	/// Cost to drop reputation changes.
 	int set_iRepdropCost = 0;
 
+	/// Cost to drop all reputation changes.
+	int set_iResetrepCost = 0;
+	// Resetrep is not allowed until server restart
+	map<string, uint> factions;
+	// Resetrep is not allowed if attempted within the resetrep time limit (in seconds)
+	int set_iResetrepTimeLimit = 0;
+
 	/// Local chat range
 	float set_iLocalChatRange = 9999;
 
@@ -72,6 +79,8 @@ namespace MiscCmds
 	{
 		// Load generic settings
 		set_iRepdropCost = IniGetI(scPluginCfgFile, "General", "RepDropCost", 0);
+		set_iResetrepCost = IniGetI(scPluginCfgFile, "General", "ResetrepCost", 10000000);
+		set_iResetrepTimeLimit = IniGetI(scPluginCfgFile, "General", "ResetrepTimeLimit", 1209600);
 		set_iLocalChatRange = IniGetF(scPluginCfgFile, "General", "LocalChatRange", 0);
 
 		set_wscStuckMsg = stows(IniGetS(scPluginCfgFile, "General", "StuckMsg", "Attention! Stand clear. Towing %player"));
@@ -79,6 +88,8 @@ namespace MiscCmds
 		set_wscCoinMsg = stows(IniGetS(scPluginCfgFile, "General", "CoinMsg", "%player tossed %result"));
 
 		set_iSmiteMusicID = CreateID(IniGetS(scPluginCfgFile, "General", "SmiteMusic", "music_danger").c_str());
+
+		LoadListOfReps();
 	}
 
 	/** Clean up when a client disconnects */
@@ -248,6 +259,185 @@ namespace MiscCmds
 		{
 			HkAddCash(wscCharname, 0 - set_iRepdropCost);
 		}
+
+		return true;
+	}
+
+	void MiscCmds::LoadListOfReps()
+	{
+		INI_Reader ini;
+
+		string factionpropfile = "..\\data\\initialworld.ini";
+		if (ini.open(factionpropfile.c_str(), false))
+		{
+			while (ini.read_header())
+			{
+				if (ini.is_header("Group"))
+				{
+					uint ids_name;
+					string nickname;
+					while (ini.read_value())
+					{
+						if (ini.is_value("nickname"))
+						{
+							nickname = ini.get_value_string();
+						}
+						else if (ini.is_value("ids_name"))
+						{
+							ids_name = ini.get_value_int(0);
+						}
+
+					}
+					factions[nickname] = ids_name;
+				}
+			}
+			ini.close();
+			ConPrint(L"Resetrep: Loaded %u factions\n", factions.size());
+		}
+	}
+	
+	map<wstring, uint> MiscCmds::Resetrep_load_Time_limits_for_player_account(string filename)
+	{
+		INI_Reader ini;
+		string factionpropfile = filename;
+		map<wstring, uint> tempmap;
+		if (ini.open(factionpropfile.c_str(), false))
+		{
+			while (ini.read_header())
+			{
+				if (ini.is_header("Playership"))
+				{
+					uint lastresettime = 0;
+					wstring nickname = L"0";
+					while (ini.read_value())
+					{
+						if (ini.is_value("shipname"))
+						{
+							ini_get_wstring(ini, nickname);
+						}
+						else if (ini.is_value("lastresettime"))
+						{
+							lastresettime = ini.get_value_int(0);
+						}
+					}
+
+					tempmap[nickname] = lastresettime;
+				}
+			}
+			ini.close();
+		}
+		return tempmap;
+	}
+
+	void MiscCmds::Resetrep_save_Time_limits_to_player_account(string filename, map<wstring, uint> tempmap)
+	{
+		string factionpropfile = filename;
+
+		FILE *file = fopen(factionpropfile.c_str(), "w");
+		if (file)
+		{
+			for (map<wstring, uint>::iterator i = tempmap.begin();
+				i != tempmap.end(); ++i)
+			{
+				if (((int)time(0) - i->second) < set_iResetrepTimeLimit)
+				{
+					wstring temp = i->first;
+					fprintf_s(file, "[Playership]\n");
+					ini_write_wstring(file, "shipname", temp);
+					fprintf_s(file, "lastresettime = %d\n", i->second);
+				}
+			}
+			fclose(file);
+		}
+	}
+
+	/** A command to reset all reputations to zero that you might have */
+	bool MiscCmds::UserCmd_ResetRep(uint iClientID, const wstring &wscCmd, const wstring &wscParam, const wchar_t *usage)
+	{
+		HK_ERROR err;
+
+		wstring wscCharname = (const wchar_t*)Players.GetActiveCharacterName(iClientID);
+
+		uint iShip;
+		pub::Player::GetShip(iClientID, iShip);
+		if (iShip)
+		{
+			PrintUserCmdText(iClientID, L"ERR Not in base");
+			return true;
+		}
+
+		// Read the current number of credits for the player
+		// and check that the character has enough cash.
+		int iCash = 0;
+		if ((err = HkGetCash(wscCharname, iCash)) != HKE_OK)
+		{
+			PrintUserCmdText(iClientID, L"ERR (1) %s", HkErrGetText(err).c_str());
+			return true;
+		}
+
+		if (set_iResetrepCost > 0 && iCash < set_iResetrepCost)
+		{
+			PrintUserCmdText(iClientID, L"ERR Insufficient credits");
+			return true;
+		}
+
+
+		wstring wscCharFileName;
+		if ((err = HkGetCharFileName(wscCharname, wscCharFileName)) != HKE_OK)
+		{
+			PrintUserCmdText(iClientID, L"ERR (2)" + HkErrGetText(err));
+			return true;
+		}
+
+		// Read the last time a rename was done on this character
+		wstring wscDir;
+		if ((err = HkGetAccountDirName(wscCharname, wscDir)) != HKE_OK)
+		{
+			PrintUserCmdText(iClientID, L"ERR (3)" + HkErrGetText(err));
+			return true;
+		}
+		string scRenameFile = scAcctPath + wstos(wscDir) + "\\" + "resetrep.ini";
+		map<wstring, uint> CharfilesLastResetRepUsage = Resetrep_load_Time_limits_for_player_account(scRenameFile);
+		if (CharfilesLastResetRepUsage.find(wscCharFileName) != CharfilesLastResetRepUsage.end())
+		{
+			uint LastTime = CharfilesLastResetRepUsage[wscCharFileName];
+
+			if (((int)time(0) - LastTime) < set_iResetrepTimeLimit)
+			{
+				int SecondsLeft = (set_iResetrepTimeLimit - ((int)time(0) - int(LastTime)));
+				int DaysLeft = SecondsLeft / (3600 * 24);
+				PrintUserCmdText(iClientID, L"ERR Attempt to resetrep too frequently. Please wait %d days (%d in seconds).", DaysLeft, SecondsLeft);
+				return true;
+			}
+		}
+
+		int count = 0;
+		for (map<string, uint>::iterator iter = factions.begin(); iter != factions.end(); iter++)
+		{
+			float fRep = 0.0f;
+			string factionName = iter->first;
+			HK_ERROR error;
+			if ((error = HkGetRep(wscCharname, stows(factionName), fRep)) != HKE_OK)
+			{
+				PrintUserCmdText(iClientID, L"ERR (5) %s", HkErrGetText(error).c_str());
+				continue;
+			}
+			count++;
+			HkSetRep(wscCharname, stows(factionName), 0.0f);
+		}
+
+		PrintUserCmdText(iClientID, L"OK Rep reset has beens performed. %d factions were cleared.", count);
+
+		// Remove cash if we're charging for it.
+		if (set_iResetrepCost > 0)
+		{
+			HkAddCash(wscCharname, 0 - set_iResetrepCost);
+		}
+
+		CharfilesLastResetRepUsage[wscCharFileName] = (int)time(0);
+		Resetrep_save_Time_limits_to_player_account(scRenameFile, CharfilesLastResetRepUsage);
+		HkSaveChar(wscCharname); //Save char just in case
+		HkKickReason(wscCharname, L"Updating character, please wait 10 seconds before reconnecting");
 
 		return true;
 	}
--- a/Plugins/Public/playercntl_plugin/MiscCmds.cpp
+++ b/Plugins/Public/playercntl_plugin/MiscCmds.cpp
@@ -66,8 +66,10 @@ namespace MiscCmds
 
 	/// Cost to drop all reputation changes.
 	int set_iResetrepCost = 0;
-	// Resetrep is not allowed until server restart
+	
+	//List of factions to be cleared by resetrep
 	map<string, uint> factions;
+
 	// Resetrep is not allowed if attempted within the resetrep time limit (in seconds)
 	int set_iResetrepTimeLimit = 0;
 
@@ -275,12 +277,12 @@ namespace MiscCmds
 				if (ini.is_header("Group"))
 				{
 					uint ids_name;
-					string nickname;
+					string faction_name;
 					while (ini.read_value())
 					{
 						if (ini.is_value("nickname"))
 						{
-							nickname = ini.get_value_string();
+							faction_name = ini.get_value_string();
 						}
 						else if (ini.is_value("ids_name"))
 						{
@@ -288,7 +290,7 @@ namespace MiscCmds
 						}
 
 					}
-					factions[nickname] = ids_name;
+					factions[faction_name] = ids_name;
 				}
 			}
 			ini.close();
@@ -299,21 +301,20 @@ namespace MiscCmds
 	map<wstring, uint> MiscCmds::Resetrep_load_Time_limits_for_player_account(string filename)
 	{
 		INI_Reader ini;
-		string factionpropfile = filename;
 		map<wstring, uint> tempmap;
-		if (ini.open(factionpropfile.c_str(), false))
+		if (ini.open(filename.c_str(), false))
 		{
 			while (ini.read_header())
 			{
 				if (ini.is_header("Playership"))
 				{
 					uint lastresettime = 0;
-					wstring nickname = L"0";
+					wstring ship_filename = L"";
 					while (ini.read_value())
 					{
 						if (ini.is_value("shipname"))
 						{
-							ini_get_wstring(ini, nickname);
+							ini_get_wstring(ini, ship_filename);
 						}
 						else if (ini.is_value("lastresettime"))
 						{
@@ -321,7 +322,7 @@ namespace MiscCmds
 						}
 					}
 
-					tempmap[nickname] = lastresettime;
+					tempmap[ship_filename] = lastresettime;
 				}
 			}
 			ini.close();
@@ -331,14 +332,14 @@ namespace MiscCmds
 
 	void MiscCmds::Resetrep_save_Time_limits_to_player_account(string filename, map<wstring, uint> tempmap)
 	{
-		string factionpropfile = filename;
 
-		FILE *file = fopen(factionpropfile.c_str(), "w");
+		FILE *file = fopen(filename.c_str(), "w");
 		if (file)
 		{
 			for (map<wstring, uint>::iterator i = tempmap.begin();
 				i != tempmap.end(); ++i)
 			{
+				////This is intended  to write to disk only time limits which have not expired yet
 				if (((int)time(0) - i->second) < set_iResetrepTimeLimit)
 				{
 					wstring temp = i->first;
@@ -371,7 +372,7 @@ namespace MiscCmds
 		int iCash = 0;
 		if ((err = HkGetCash(wscCharname, iCash)) != HKE_OK)
 		{
-			PrintUserCmdText(iClientID, L"ERR (1) %s", HkErrGetText(err).c_str());
+			PrintUserCmdText(iClientID, L"ERR Resetrep failed to read player cash");
 			return true;
 		}
 
@@ -385,7 +386,7 @@ namespace MiscCmds
 		wstring wscCharFileName;
 		if ((err = HkGetCharFileName(wscCharname, wscCharFileName)) != HKE_OK)
 		{
-			PrintUserCmdText(iClientID, L"ERR (2)" + HkErrGetText(err));
+			PrintUserCmdText(iClientID, L"ERR Resetrep failed to get charfilename");
 			return true;
 		}
 
@@ -393,7 +394,7 @@ namespace MiscCmds
 		wstring wscDir;
 		if ((err = HkGetAccountDirName(wscCharname, wscDir)) != HKE_OK)
 		{
-			PrintUserCmdText(iClientID, L"ERR (3)" + HkErrGetText(err));
+			PrintUserCmdText(iClientID, L"ERR Resetrep failed to get account dir");
 			return true;
 		}
 		string scRenameFile = scAcctPath + wstos(wscDir) + "\\" + "resetrep.ini";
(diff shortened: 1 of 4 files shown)


PhoenixxClaw:
First of all, sorry for jumping in here. :D

I'd recommend you to recode this a bit @dd84ai ... If I understand your code correctly you're basically setting the reputation of all currently existing factions to 0.0 with 'HkSetRep(wscCharname, stows(factionName), 0.0f);'. Problem with that is that you'd set this for factions who are supposed to be hostile as well. (means a Nomad IDed Character would suddenly be able to be neutral to all factions).  

I recommend adding a check for the currently equipped ID of the ship, get the default rep for it and set the rep to the defaults.

RnScully:
I think a way to reset rep would be very helpful to many players. Changing IDs can get you into very weird places.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +264,178 @@ namespace MiscCmds
...
+					while (ini.read_value())
+					{
+						if (ini.is_value("shipname"))
+						{
+							ini_get_wstring(ini, nickname);
+						}
+						else if (ini.is_value("timelimit"))
+						{
+							ids_name = ini.get_value_int(0);
+							ShipnamesWhichAlreadyUsedResetRep[nickname] = ids_name;
```
dsyalex, line 324: this should be done after the `while` block, it will otherwise break if someone tries to specify  `timelimit` before `shipname`

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +264,178 @@ namespace MiscCmds
...
+
+		if (ShipnamesWhichAlreadyUsedResetRep.find(wscCharFileName) != ShipnamesWhichAlreadyUsedResetRep.end())
+		{
+			uint LastTime = ShipnamesWhichAlreadyUsedResetRep[wscCharFileName];
+
+			if (((int)time(0) - LastTime) < set_iResetrepTimeLimit)
+			{
+				int SecondsLeft = (set_iResetrepTimeLimit - ((int)time(0) - int(LastTime)));
+				int DaysLeft = SecondsLeft / (3600 * 24);
+			PrintUserCmdText(iClientID, L"ERR Resetrep time limit. %d days (%d in seconds) time remains to wait.", DaysLeft, SecondsLeft);
```
dsyalex, line 403: broken indentation. also this could be worded better

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +264,178 @@ namespace MiscCmds
...
+			}
+		}
+
+		int count = 0;
+		for (map<string, uint>::iterator iter = factions.begin(); iter != factions.end(); iter++)
+		{
+			float fRep = 0.0f;
+			string factionName = iter->first;
+			HK_ERROR error;
+			if ((error = HkGetRep(wscCharname, stows(factionName), fRep)) != HKE_OK)
```
dsyalex, line 414: what purpose does all this serve exactly?
do we want to check fRep != 0.0f before count++?

Plugins/Public/playercntl_plugin/Rename.cpp
```diff
@@ -455,6 +458,15 @@ namespace Rename
 
 				// The rename worked. Log it and save the rename time.
 				AddLog("NOTICE: User rename %s to %s (%s)", wstos(o.wscCharname).c_str(), wstos(o.wscNewCharname).c_str(), wstos(HkGetAccountID(acc)).c_str());
+
+				wstring wscCharFileName2;
+				HkGetCharFileName(o.wscNewCharname, wscCharFileName2);
+				//Anti renameme, for /resetrep cooldown
+				if (MiscCmds::ShipnamesWhichAlreadyUsedResetRep.find(wscCharFileName) != MiscCmds::ShipnamesWhichAlreadyUsedResetRep.end())
```
dsyalex, line 465: So we're actually using char file names in this map instead of char names?

dsyalex (changes requested):
Storing a central DB of state like this in flhook_plugins feels wrong. Don't we normally store per-char file info in files in the relevant account directory and move those around?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: Just a check for possible errors. To verify that every used factionName belongs to some registered faction.
If it would not pass, usage of HkSetRep can be not recommended... for more stable command work probably will be better having "continue;"  instead of "return true" in error catching if construction. Changed to "continue" to skip current for loop step.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: fixed

Plugins/Public/playercntl_plugin/Rename.cpp
dd84ai: yup, more reliable. char file names are having always names like abdfbd_dfdsf
only English alphabet with one "_"
I feel it is more secure to use than all the random things a regular player name can be.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: fixed indentation. no idea how to reword for now.

dd84ai:
> Storing a central DB of state like this in flhook_plugins feels wrong. Don't we normally store per-char file info in files in the relevant account directory and move those around?

yeah, you right, it would be much better. I just... presumed that central DB is good to avoid /movechar cooldown removal. But then I realized that I just can update my cooldown in during movechar operation. Rewriting.

dd84ai:
in progres.... not done yet

dd84ai:
Rewritten to decentralize database. Data is stored in account folders now.

dd84ai:
https://discoverygc.com/forums/showthread.php?tid=183637
gathered opinions

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: `PrintUserCmdText(iClientID, L"ERR Attempt to resetrep too frequently. Please wait %d days (%d in seconds).", DaysLeft, SecondsLeft);`

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: Given that this is data from initialworld isn't this effectively dead code that can never be hit?

Plugins/Public/playercntl_plugin/Rename.cpp
dsyalex: okay so we should name it that way

dsyalex:
Thanks, replies above

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: fixed

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: Yes, but.... moderators can mess up this data with being still working
This check makes sure that "SetRep" is not applied to entity which can't be changed
saving the game from possible crash to just skipped loop cycle.

This check is used very rarely anyway, only when player uses the command
So it should be as crash-proof as possible in my opinion.

Plugins/Public/playercntl_plugin/Rename.cpp
dd84ai: it was already named as CharFileName. Not sure what you mean here

dd84ai:
Everything is replied

Plugins/Public/playercntl_plugin/Rename.cpp
dsyalex: Right now this is named 'shipnames' (i.e., character names), but actually it's char file names. Not the same thing.

dsyalex:
another comment above

Plugins/Public/playercntl_plugin/Rename.cpp
dd84ai: Uh. I guess I was blind to it. Fixed.
https://github.com/DiscoveryGC/FLHook/pull/138/commits/fcc87ffc1733218a9972786c512407c5ca6b9ed9

dd84ai:
Resolved

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -64,6 +64,13 @@ namespace MiscCmds
 	/// Cost to drop reputation changes.
 	int set_iRepdropCost = 0;
 
+	/// Cost to drop all reputation changes.
+	int set_iResetdropCost = 0;
```
dsyalex, line 68: set_iResetrepCost

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -64,6 +64,13 @@ namespace MiscCmds
...
+	// Resetrep is not allowed until server restart
+	map<string, uint> factions;
```
dsyalex, line 70: `factions` is a really misleading name for this

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,185 @@ namespace MiscCmds
...
+			}
+			ini.close();
+			ConPrint(L"Resetrep: Loaded %u factions\n", factions.size());
+		}
+	}
+	
+	map<wstring, uint> MiscCmds::Resetrep_Time_limits_load_map(string filename)
+	{
+		INI_Reader ini;
+		string factionpropfile = filename;
```
dsyalex, line 302: this doesn't seem to be necessary

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,185 @@ namespace MiscCmds
...
+				}
+			}
+			ini.close();
+		}
+		return tempmap;
+	}
+
+	void MiscCmds::Resetrep_Time_limits_add(string filename, map<wstring, uint> tempmap)
+	{
+		string factionpropfile = filename;
```
dsyalex, line 334: this doesn't seem to be necessary

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,185 @@ namespace MiscCmds
...
+				{
+					uint timelimit = 0;
+					wstring nickname = L"0";
+					while (ini.read_value())
+					{
+						if (ini.is_value("shipname"))
+						{
+							ini_get_wstring(ini, nickname);
+						}
+						else if (ini.is_value("timelimit"))
```
dsyalex, line 318: shouldn't this be named more like `lastresettime`?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,185 @@ namespace MiscCmds
...
+	{
+		string factionpropfile = filename;
+
+		FILE *file = fopen(factionpropfile.c_str(), "w");
+		if (file)
+		{
+			for (map<wstring, uint>::iterator i = tempmap.begin();
+				i != tempmap.end(); ++i)
+			{
+				if (((int)time(0) - i->second) < set_iResetrepTimeLimit)
```
dsyalex, line 342: wait why is this here??

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,185 @@ namespace MiscCmds
...
+			if ((error = HkGetRep(wscCharname, stows(factionName), fRep)) != HKE_OK)
+			{
+				PrintUserCmdText(iClientID, L"ERR (5) %s", HkErrGetText(error).c_str());
+				continue;
+			}
+			count++;
+			HkSetRep(wscCharname, stows(factionName), 0.0f);
+		}
+
+		PrintUserCmdText(iClientID, L"OK Reset Rep is performed. %d factions were cleared.", count);
```
dsyalex, line 429:
```suggestion
		PrintUserCmdText(iClientID, L"OK Rep reset has beens performed. %d factions were cleared.", count);
```

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,185 @@ namespace MiscCmds
...
+
+		// Remove cash if we're charging for it.
+		if (set_iResetdropCost > 0)
+		{
+			HkAddCash(wscCharname, 0 - set_iResetdropCost);
+		}
+
+		ShipnamesWhichAlreadyUsedResetRep[wscCharFileName] = (int)time(0);
```
dsyalex, line 437: maybe name it CharfilesLastResetRepUsage?

Plugins/Public/playercntl_plugin/Rename.cpp
```diff
@@ -455,6 +463,26 @@ namespace Rename
...
+				HkGetCharFileName(o.wscNewCharname, wscCharFileName2);
+				wstring wscDir;	
+				HK_ERROR err;
+				if ((err = HkGetAccountDirName(o.wscNewCharname, wscDir)) != HKE_OK)
+				{
+					ConPrint(L"ERR 461 " + HkErrGetText(err));
+				}
+				else
+				{
+					string scRenameFile = scAcctPath + wstos(wscDir) + "\\" + "resetrep.ini";
```
dsyalex, line 478:
```suggestion
					string scRenameFile = scAcctPath + wstos(wscDir) + "\\resetrep.ini";
```

Plugins/Public/playercntl_plugin/Rename.cpp
```diff
@@ -873,6 +912,8 @@ namespace Rename
 		o.scSourceFile = scAcctPath + wstos(wscSourceDir) + "\\" + wstos(wscSourceFile) + ".fl";
 		o.scDestFile = scAcctPath + wstos(wscDir) + "\\" + wstos(wscSourceFile) + ".fl";
 		o.scDestFileTemp = scAcctPath + wstos(wscDir) + "\\" + wstos(wscSourceFile) + ".fl.moving";
+		o.scSourceDir = scAcctPath + wstos(wscSourceDir) + "\\" + "resetrep.ini";
```
dsyalex, line 915:
```suggestion
		o.scSourceDir = scAcctPath + wstos(wscSourceDir) + "\\resetrep.ini";
```

Plugins/Public/playercntl_plugin/Rename.cpp
```diff
@@ -873,6 +912,8 @@ namespace Rename
...
+		o.scDestDir = scAcctPath + wstos(wscDir) + "\\" + "resetrep.ini";
```
dsyalex, line 916:
```suggestion
		o.scDestDir = scAcctPath + wstos(wscDir) + "\\resetrep.ini";
```

dsyalex (changes requested):

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: if only I could remember...

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: it looks like a function that load 'cooldown restrictions' for certain player folder/account

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: it looks like a function that rewrites/saves player time limits for all present charfilenames in some player folder

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: from what I was able to find... in the code which I certainly do not remember anymore

those two extra functions were made as 'wrappers' to deal with time limits being stored in files.

it looks to be very useful when 'data' about time limits needs to be 'loaded' again in rename and move char functions
image

dd84ai: more precide namings for those functions would be
Resetrep_load_time_limit_for_player account(folder)
Resetrep_save_time_limits_for_player_account(folder)

dd84ai: After re-reading this code, I still think those functions are still needed. Because I don't want to write fully code to 'read' data from files for all three times it is needed

dd84ai: I could make it more looking like a one class wrapping those dealings though.

something like
class time_limits_for_player_account
{
map_of_time_limits
construct()
{ load from file }
save()
{ save to file }
get_time_limits()
{ return map_of_time_limits }
}

dd84ai: Resetrep_save_time_limits_for_player_account is used three times, good enough to be a function and not copy pasted code

dd84ai: commited and squashed renames to functions

dd84ai:
answered

dd84ai:
Doing tests:
1) I had hostile reps, did /resetrep, was kicked, reps aren't hostile anymore
2) Attempt to do second /resetrep, denied and informed to wait 13 days (1209538 seconds)
2) Attempt to do third /resetrep, denied and informed to wait 13 days (1209501 seconds)
3) Edited by server side files to change last time of resetrep to 2'000'000 seconds into the past, attempted resetrep second time, succeded
4) ....test fail.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: right but why not just either rename `filename` to `factionpropfile`, or drop all references to `factionpropfile` and use `filename`?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: see above

dsyalex (changes requested):

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: indeed.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: understood the issue

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -64,6 +64,13 @@ namespace MiscCmds
 	/// Cost to drop reputation changes.
 	int set_iRepdropCost = 0;
 
+	/// Cost to drop all reputation changes.
+	int set_iResetrepCost = 0;
+	// Resetrep is not allowed until server restart
```
dsyalex, line 69: this doc comment is not related to the variable below it

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,183 @@ namespace MiscCmds
...
+		INI_Reader ini;
+		map<wstring, uint> tempmap;
+		if (ini.open(filename.c_str(), false))
+		{
+			while (ini.read_header())
+			{
+				if (ini.is_header("Playership"))
+				{
+					uint lastresettime = 0;
+					wstring nickname = L"0";
```
dsyalex, line 310: weird default ship name? nickname is also a weird name.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,183 @@ namespace MiscCmds
...
+		if (file)
+		{
+			for (map<wstring, uint>::iterator i = tempmap.begin();
+				i != tempmap.end(); ++i)
+			{
+				if (((int)time(0) - i->second) < set_iResetrepTimeLimit)
+				{
+					wstring temp = i->first;
+					fprintf_s(file, "[Playership]\n");
+					ini_write_wstring(file, "shipname", temp);
```
dsyalex, line 344: directly use i->first ?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dsyalex: I still don't understand why we're checking the current time inside a function intended to write a map to disk.

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,183 @@ namespace MiscCmds
...
+		{
+			PrintUserCmdText(iClientID, L"ERR Insufficient credits");
+			return true;
+		}
+
+
+		wstring wscCharFileName;
+		if ((err = HkGetCharFileName(wscCharname, wscCharFileName)) != HKE_OK)
+		{
+			PrintUserCmdText(iClientID, L"ERR (2)" + HkErrGetText(err));
```
dsyalex, line 386: why do we concatenate here but do string formatting and c_str above? also missing space after `(2)`?

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,183 @@ namespace MiscCmds
...
+			return true;
+		}
+
+		// Read the last time a rename was done on this character
+		wstring wscDir;
+		if ((err = HkGetAccountDirName(wscCharname, wscDir)) != HKE_OK)
+		{
+			PrintUserCmdText(iClientID, L"ERR (3)" + HkErrGetText(err));
```
dsyalex, line 394: see above

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,183 @@ namespace MiscCmds
...
+			if ((error = HkGetRep(wscCharname, stows(factionName), fRep)) != HKE_OK)
+			{
+				PrintUserCmdText(iClientID, L"ERR (5) %s", HkErrGetText(error).c_str());
+				continue;
+			}
+			count++;
+			HkSetRep(wscCharname, stows(factionName), 0.0f);
+		}
+
+		PrintUserCmdText(iClientID, L"OK Rep reset has beens performed. %d factions were cleared.", count);
```
dsyalex, line 427:
```suggestion
		PrintUserCmdText(iClientID, L"OK Rep reset has been performed. %d factions were cleared.", count);
```

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +263,183 @@ namespace MiscCmds
...
+
+		int count = 0;
+		for (map<string, uint>::iterator iter = factions.begin(); iter != factions.end(); iter++)
+		{
+			float fRep = 0.0f;
+			string factionName = iter->first;
+			HK_ERROR error;
+			if ((error = HkGetRep(wscCharname, stows(factionName), fRep)) != HKE_OK)
+			{
+				PrintUserCmdText(iClientID, L"ERR (5) %s", HkErrGetText(error).c_str());
```
dsyalex, line 420: we missed `(4)`?

Plugins/Public/playercntl_plugin/Rename.cpp
```diff
@@ -455,6 +463,26 @@ namespace Rename
...
 				AddLog("NOTICE: User rename %s to %s (%s)", wstos(o.wscCharname).c_str(), wstos(o.wscNewCharname).c_str(), wstos(HkGetAccountID(acc)).c_str());
+
+				//resetrep cooldown
+				wstring wscCharFileName2;
+				HkGetCharFileName(o.wscNewCharname, wscCharFileName2);
+				wstring wscDir;	
+				HK_ERROR err;
+				if ((err = HkGetAccountDirName(o.wscNewCharname, wscDir)) != HKE_OK)
+				{
+					ConPrint(L"ERR 461 " + HkErrGetText(err));
```
dsyalex, line 474: okay we could do something better than 461 here if it's printing errors to console - like the function name and that it's an HkGetAccountDirName check for resetrep. But also what about logging to disk?

dsyalex (changes requested):

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: ship_filename willl be better I guess

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: Sure

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: >I still don't understand why we're checking the current time inside a function intended to write a map to disk.
This was intended... to write to disk only time limits which have not expired yet

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: renamed

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: renamed error

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: Ok

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: Renamed

Plugins/Public/playercntl_plugin/Rename.cpp
dd84ai: Uh, they aren't ever should happen at all. So no need to log

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: Fixed!

dd84ai:
`I still don't understand why we're checking the current time inside a function intended to write a map to disk.`
This was intended... to write to disk only time limits which have not expired yet

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: This was intended... to write to disk only time limits which have not expired yet
nvm, answering in github

dd84ai:
`I still don't understand why we're checking the current time inside a function intended to write a map to disk.`

This was intended... to write to disk only time limits which have not expired yet

Plugins/Public/playercntl_plugin/MiscCmds.cpp
```diff
@@ -252,6 +265,184 @@ namespace MiscCmds
...
+			float fRep = 0.0f;
+			string factionName = iter->first;
+			HK_ERROR error;
+			if ((error = HkGetRep(wscCharname, stows(factionName), fRep)) != HKE_OK)
+			{
+				PrintUserCmdText(iClientID, L"ERR Resetrep, check for faction failed");
+				continue;
+			}
+			count++;
+			HkSetRep(wscCharname, stows(factionName), 0.0f);
```
dsyalex, line 427: just sets all factions to neutral? shall we leave a comment saying setting all factions to neutral is safe because the player must be docked and rep hacks will take effect on undock?

Plugins/Public/playercntl_plugin/Rename.cpp
```diff
@@ -455,6 +463,26 @@ namespace Rename
...
 				AddLog("NOTICE: User rename %s to %s (%s)", wstos(o.wscCharname).c_str(), wstos(o.wscNewCharname).c_str(), wstos(HkGetAccountID(acc)).c_str());
+
+				//resetrep cooldown
+				wstring wscCharFileName2;
+				HkGetCharFileName(o.wscNewCharname, wscCharFileName2);
+				wstring wscDir;	
+				HK_ERROR err;
+				if ((err = HkGetAccountDirName(o.wscNewCharname, wscDir)) != HKE_OK)
+				{
+					ConPrint(L"ERR Resetrep failure in rename to get directory");
```
dsyalex, line 474: missing newline?

dsyalex (changes requested):
almost there

Plugins/Public/playercntl_plugin/Rename.cpp
dd84ai: Right

Plugins/Public/playercntl_plugin/MiscCmds.cpp
dd84ai: All right.

dd84ai: Added
//We reset reps to zero, but it is safe because the ship is docked and kicked
//So rephacks take effect right after logging back

dsyalex (approved):

PR #139, opened 2020-09-30
base: remove hostile tagged players from base to proxy

dd84ai:
This change makes sure that players who were added to /base addhostile
Will be exiting base at the proxy (100k below sun).
Their access to the base will be stripped fully when they exit base or log out of the game, changing their resurrecting point to proxy base.

What is it for? For situations when pirates dock neutral to them pob and begin hunting base owners.
And just adding to the hostile list is not enough, since they can suicide to return back. It prevents it.

Sort of delayed 'kick' from a base, making sure that hostile list works better.


dd84ai:
Closing because detected incorrect work

dd84ai:
Tbh, it is dealing with too many bugs. I do not wish to continue fighting problems raised with it.

PR #140, opened 2020-10-06
mobiledocking: Docking Module Massive Upgrade (12.10.12)

dd84ai:
Full changelog/description here: https://discoverygc.com/forums/showthread.php?tid=183746&pid=2193999#pid2193999
Shortly speaking it is full docking module rework

Main feature additions:
(+) Adding the ability for carried by docking module snubs to disconnect and log later with appearance still inside.
(+) Adding the ability for a carrier to relog and still being keeping snubs (snubs can undock with offline carrier also)
(+) Adding the ability to dock not-in-game capital vessel which is at the same base.
(+) Snubs being kicked by carrier in some base, get transferred themselves to the base it is in.
(+) You can buy ammo and batts inside the carrier now! (Minus cloak/jump batteries)(Items are copied from Green Station)
(+) Replaced all messy old commands with new system of commands

Many bugs were fixed:
(+) New command [color=#00FF00]/dock send[/color] to have stable and reliable alternative to dock (F3 is still a choice)
(+) Distances for dock and allowdock are calculated in a right way(considering object radiuses) now
(+) Fixed lack of compatibility when plugin interacts with POB instead of normal NPC base.
(+) Notify players around about docking with TRAFFIC CONTROL
(+) Fixed bug with wrong module available amount calculation after hot plugin restart
(+) fixed to have dock kick by number, not name.
(+) Fighter which docks with cargo inside of the fighter, will be 'undocked' to proxy base if it disconnects from the game. (to prevent offline way to increase cargo for smuggling)


roalyr:
```mapInviteLinks[iBaseID].Password = index;```

If two carriers issue the same invite, wouldn't the first one be overridden? Maybe add a vector that stores those invites and prevents duplicates? Also, maybe the time limit should be increased (re-logging from different accounts?)

dd84ai:
> `mapInviteLinks[iBaseID].Password = index;`
> 
> If two carriers issue the same invite, wouldn't the first one be overridden? Maybe add a vector that stores those invites and prevents issuing the same invites until they are expired? Also, maybe the time limit should be increased (re-logging from different accounts?)

Good notice. I saw it too, but forgot about it.
As for time limit... I think it can be increased as maximum to two minutes, or better solving problem how to check that offline carrier did not leave the base. That is why I have tight time limit expiration just in case.

dd84ai:
Fixed issues mentioned above and some other noticed ones

roalyr:
What about carrying potentially hostile vessel? Smugglin one's own ships like trojan horse.

roalyr:
Get contraband in freighters, dock onto neutral carrier, haul stuff.

roalyr:
Maybe some rephack restrictions, or better, an ability to expose cargo within hauled ships.

dd84ai:
> Maybe some rephack restrictions, or better, an ability to expose cargo within hauled ships.

I don't want to bother with rephack restrictions, it will make it too restrictive for roleplay reasons.
On other hands I see no problem in forbidding having cargo. I can just copy whatever check there is for command '/conn'

dd84ai:
Uh, something is wrong here.

dd84ai:
Some part of code related to "Penalty to snubs carried in offline way with cargo" works really bad. I'll try to analyze which part and then to fix it.

dd84ai:
Canceling alert. It looks like a bad idea to use clients and servers of different versions. That's the reason for my failures.
A small check for a minor thing is still required though.

dsyalex (commented):
Needs rebase (also please note that given the size of this and the fact it's about mobiledocking rather than plugins we have live, the priority is lower on this than the other PRs)

dd84ai:
> Needs rebase (also please note that given the size of this and the fact it's about mobiledocking rather than plugins we have live, the priority is lower on this than the other PRs)

>  **about mobiledocking rather than plugins we have live**

aren't we having it working on server at the moment? It is named dock.dll instead of mobiledock.dll possibly.

dd84ai:
Updated to be up to date with master (since there araised merging issues due to being outdated)

dsyalex:
> > Needs rebase (also please note that given the size of this and the fact it's about mobiledocking rather than plugins we have live, the priority is lower on this than the other PRs)
> 
> > **about mobiledocking rather than plugins we have live**
> 
> aren't we having it working on server at the moment? It is named dock.dll instead of mobiledock.dll possibly.

Oof good point. Now I'm even more hesitant to approve rewrites of it though, especially given historical problems with this plugin :)

dd84ai:
> > > Needs rebase (also please note that given the size of this and the fact it's about mobiledocking rather than plugins we have live, the priority is lower on this than the other PRs)
> > 
> > 
> > > **about mobiledocking rather than plugins we have live**
> > 
> > 
> > aren't we having it working on server at the moment? It is named dock.dll instead of mobiledock.dll possibly.
> 
> Oof good point. Now I'm even more hesitant to approve rewrites of it though, especially given historical problems with this plugin :)

I had no major problems during rework
The worst situation that happens with a player, he gets thrown to 'the solar base which is 100k below sun' and he has to fly back on his own.

PR #120, opened 2020-09-03
Base: Triggers weapon platforms from sum of damage

dd84ai:
This change makes neutral or friendly aligned weapon platforms triggered to hostile state not just from any damage 
but from certain sum of damage performed by one player during 24 hours
for example, 200000 (float setting damage_treshold is added for config file)

Every player has separate counter remembering how much damage they did to base (attached to their ship name)

PR #144, opened 2021-01-02
base: weapon platform hostility delay.

dd84ai:
Creating to replace https://github.com/DiscoveryGC/FLHook/pull/120 which was broken due to force pushed commits. Apperently I should re open PR before force pushing commits, if I don't want to close PR forever.

Description:
This change makes neutral or friendly aligned weapon platforms triggered to hostile state not just from any damage
but from certain sum of damage performed by one player during 24 hours
for example, 400000 (float setting damage_threshold is added for config file)
Every player has separate counter remembering how much damage they did to base (attached to their ship name)

### DIFF
--- a/Plugins/Public/base_plugin/PlayerBase.cpp
+++ b/Plugins/Public/base_plugin/PlayerBase.cpp
@@ -677,7 +677,15 @@ float PlayerBase::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 			{
 				if (set_plugin_debug > 1)
 					ConPrint(L"PlayerBase::damaged space_obj=%u\n", space_obj);
-				hostile_tags[charname] = charname;
+				
+				if (hostile_tags_damage.find(charname) == hostile_tags_damage.end())
+					hostile_tags_damage[charname] = 0;
+
+				if ((hostile_tags_damage[charname] + damage) < damage_threshold)
+					hostile_tags_damage[charname] += damage;
+				else hostile_tags[charname] = charname;
+					
+
 				SyncReputationForBase();
 			}
 		}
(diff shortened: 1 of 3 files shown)


dd84ai:
good commit descriptions for everyone!

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -677,7 +677,16 @@ float PlayerBase::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 			{
 				if (set_plugin_debug > 1)
 					ConPrint(L"PlayerBase::damaged space_obj=%u\n", space_obj);
-				hostile_tags[charname] = charname;
+				
+				if (hostile_tags_damage.find(charname) != hostile_tags_damage.end())
+				{
+					if (hostile_tags_damage[charname] < damage_threshold)
```
dsyalex, line 683:
```suggestion
					if ((hostile_tags_damage[charname] + damage) < damage_threshold)
```

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -677,7 +677,16 @@ float PlayerBase::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
...
+						hostile_tags_damage[charname] += damage;
+					else hostile_tags[charname] = charname;
+				}
+				else hostile_tags_damage[charname] = damage;
```
dsyalex, line 687: this doesn't account for the first shot being over the threshold?

dsyalex (changes requested):

Plugins/Public/base_plugin/PlayerBase.cpp
dd84ai: It accounts now

dsyalex (approved):

PR #147, opened 2021-01-31
Weapon platform changes

dd84ai:
fix(base): fixing incorrect work of commit which adds damage_treshlold to trigger weapon platform hostility


dsyalex (commented):
(done in #148)

PR #163, opened 2022-04-16
Feature: Added Buy/Sell Functionality

Zentor88:
**Feature: Added Buy/Sell Functionality**
Added sellprice variable to handle the logic and updated the transaction code to accomplish.
Freelancer only supports one value displayed in dealer menu's however, so the /price command was added. 
Available to all visitors of the base, the /price buy changes the prices displayed in the menu to what the base is buying from players for, /price sell changes it to what the base is selling for. There is condition statements in place to not reveal all items listed in the /shop menu when the /price commands are used
Also changed dealer view logic to display what the base buys the item for if the item you have is out of stock on the base, otherwise by default, prices shown are what the base sells the item for.

_/base help menus updated with feature info_

### DIFF
--- a/Plugins/Public/base_plugin/ClientCommands.cpp
+++ b/Plugins/Public/base_plugin/ClientCommands.cpp
@@ -38,14 +38,14 @@ void SendMarketGoodUpdated(PlayerBase *base, uint good, MARKET_ITEM &item)
 				if (item.quantity == 0)
 				{
 					_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-						base->proxy_base, good, item.price, 1, 0);
+						base->proxy_base, good, item.buyprice, 1, 0);
 				}
 				// If the item is buy only and this is not an admin then it is
 				// buy only at the client
 				else if (item.min_stock >= item.quantity && !clients[client].admin)
 				{
 					_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-						base->proxy_base, good, item.price, 1, 0);
+						base->proxy_base, good, item.buyprice, 1, 0);
 				}
 				// Otherwise this item is for sale by the client.
 				else
@@ -80,14 +80,14 @@ void SendMarketGoodSync(PlayerBase *base, uint client)
 		if (item.quantity == 0)
 		{
 			_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-				base->proxy_base, good, item.price, 1, 0);
+				base->proxy_base, good, item.buyprice, 1, 0);
 		}
 		// If the item is buy only and this is not an admin then it is
 		// buy only at the client
 		else if (item.min_stock >= item.quantity && !clients[client].admin)
 		{
 			_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-				base->proxy_base, good, item.price, 1, 0);
+				base->proxy_base, good, item.buyprice, 1, 0);
 		}
 		// Otherwise this item is for sale by the client.
 		else
--- a/Plugins/Public/base_plugin/ClientCommands.cpp
+++ b/Plugins/Public/base_plugin/ClientCommands.cpp
@@ -38,20 +38,20 @@ void SendMarketGoodUpdated(PlayerBase *base, uint good, MARKET_ITEM &item)
 				if (item.quantity == 0)
 				{
 					_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-						base->proxy_base, good, item.buyprice, 1, 0);
+						base->proxy_base, good, item.buyprice, 2, 0);
 				}
 				// If the item is buy only and this is not an admin then it is
 				// buy only at the client
 				else if (item.min_stock >= item.quantity && !clients[client].admin)
 				{
 					_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-						base->proxy_base, good, item.buyprice, 1, 0);
+						base->proxy_base, good, item.buyprice, 2, 0);
 				}
 				// Otherwise this item is for sale by the client.
 				else
 				{
 					_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-						base->proxy_base, good, item.sellprice, 0, item.quantity);
+						base->proxy_base, good, item.sellprice, 2, item.quantity);
 				}
 				SendCommand(client, buf);
 			}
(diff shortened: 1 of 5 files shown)


Plugins/Public/base_plugin/ClientCommands.cpp
```diff
@@ -51,7 +51,7 @@ void SendMarketGoodUpdated(PlayerBase *base, uint good, MARKET_ITEM &item)
 				else
 				{
 					_snwprintf(buf, sizeof(buf), L" SetMarketOverride %u %u %f %u %u",
-						base->proxy_base, good, item.price, 0, item.quantity);
```
dd84ai, line 54: How about renaming item.price to more appropriate item.buyprice then?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1669,7 +1675,7 @@ namespace PlayerCommands
 			}
 		}
 
-		int pages = (matchingItems / 40) + 1;
+		int pages = (matchingItems / 30) + 1;
```
dd84ai, line 1678: 30 looks like a magical number. Perhaps to rename to what is it in reality
int ItemsPerPage = 30;
and to use the variable in the formulas?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1736,6 +1742,65 @@ namespace PlayerCommands
 		pub::Player::PopUpDialog(client, caption, message, POPUPDIALOG_BUTTONS_CENTER_OK);
 	}
 
+	//FL only supports one "price" in the dealer menu, this function gives players a command to
```
dd84ai, line 1745: What happens when the player buys item, when the view shows not the right view? Have you tried doing it?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1736,6 +1742,65 @@ namespace PlayerCommands
...
+			for (map<uint, MARKET_ITEM>::iterator i = base->market_items.begin();
+				i != base->market_items.end(); i++)
+			{
+				uint good = i->first;
+				MARKET_ITEM& item = i->second;
+				wchar_t buf[200];
+
+				//this if statement prevents items intended to be hidden from being displayed
+				//when the /pice commands run
+				if (item.min_stock > 2000 || (item.min_stock == item.max_stock)) {
```
dd84ai, line 1770: Why this condition to hide if min_stock > 2000?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1774,10 +1841,11 @@ namespace PlayerCommands
 					i->second.price = (float)money;
 					i->second.min_stock = min_stock;
 					i->second.max_stock = max_stock;
+					i->second.sellprice = (float)sellmoney;
 					SendMarketGoodUpdated(base, i->first, i->second);
 					base->Save();
-
-					int page = ((curr_item + 39) / 40);
+					//lowering how much appears on a page to keep the card from bugging out with new sell content
+					int page = ((curr_item + 29) / 30);
```
dd84ai, line 1848: Can we turn those magical numbers into less magical numbers too.

Plugins/Public/base_plugin/PlayerCommands.cpp
Zentor88: The dealer view will not always show the correct price without use of the /price command but will handle the transaction correctly. There is no way I can find that make FL change the displayed value when only clicked upon. So if a player is making a purchase and use /price sell the view will show correct values.

Plugins/Public/base_plugin/PlayerCommands.cpp
Zentor88: This was in attempt to invis stock that hasn't been configured in /shop price menu and from looking at some examples of live bases, to handle coniditions where owners use arbitrary high number to hide supply stock

Plugins/Public/base_plugin/PlayerCommands.cpp
dd84ai: I would imagine it would add some confusion to players.
Imagine the player visits the base and it has Buy prices mod. He happily starts to buy for the price he thinks the base is selling one.
And he discovers it is much higher than the buying price. Perhaps to give him in some way some warning at least?
Or may be even canceling the purchase if he is not in the right mod?

Plugins/Public/base_plugin/PlayerCommands.cpp
dd84ai: If I remember correctly the default behaviour:
The players can see the prices of items which player base is willing to sell (if item.count > min_stock), or willing to buy (if item.count < max_stock)
Plus player can see all shop prices if he has guest base password.

What's exactly wrong with the current permission system, and why not to use it again? Why it needs to have item invisibility added to it with additional conditions?

dd84ai: The player hide the stock of item with ridiculous high min_stock value, you are right.
And lets say it is the base with Ore. Which can have up to... 100 or 200k thousands of Ores.
So.... the min_stock > 2000 would be not really useful to hide it at all.

High min_stock forbids from selling (if min_stock is higher than item count, the base will not sell anything)
Zero max_stock forbids from buying. (if max_stock 0, the base will not buy anyithing)

What does min_stock == max_stock do then?

Zentor88: I'll start by answering the easy part: min_stock == max_stock condition is there because by default, items are loaded into the /shop as min=100000 max=100000, so setting that condition will auto hide those items until owner/manager configures what he wants them to be. Also to disable buy/sell, currently players can set min and max to 0, so it takes care of keeping these items invis'ed again.

upon looking further, the min_stock > 2000 does seem redundant and not of use, though I thought I had a reason when configuring it. (need to keep better comments it seems).

Plugins/Public/base_plugin/PlayerCommands.cpp
Zentor88: You raise a good point, I put a contingency in place: I changed the market display  logic (such that I could) so that when you first enter the market it displays the bases selling prices, unless you open the market with something on hand that the base has 0 of, in which case it displays what the base buys it for

Plugins/Public/base_plugin/ClientCommands.cpp
Zentor88: yes that makes sense. We can do this

Plugins/Public/base_plugin/PlayerCommands.cpp
dd84ai: > Also to disable buy/sell, currently players can set min and max to 0, so it takes care of keeping these items invis'ed again.

I will repeat again. When player sets min and max to 0. He sets to sell all of his items. It is perfectly valid selling mod for the item.

Zentor88: yes you're right, it does allow selling of remaining stock, in current set up, owners keep items in this state and sell ore to the base while logged in to override these values and not have to switch them, essentially keeping base in sell mode. But that is based on the limitation of the current set up. With the ability to have separate buy and sell prices, base owners will only use this min 0, max 0 if they want to shut down both buying and selling and therefore won't need to show their prices.

As stated this conditional is there because with the /price command it reveals all set prices of everything listed in /shop, but with the conditional in place, it only shows relevant values.

Zentor88:
Implemented suggested changes except the min/max conditional as it does have use in the /price command

PR #166, opened 2022-12-10
Siege Rework

Aingar:
Reworked the shield system to the following:

Base.cfg specifies 'vulnerability windows', time periods where bases are vulnerable. Outside of those windows, bases are straight up impervious to damage other than wear&tear. 

Shield damage reduction starts relatively low, and ramps up the reduction as station takes damage, possibly going up to 100%, rendering the station invincible before the vulnerability period ends.

The damage reduction goes back to the specified default upon entering a vulnerability window.

Config file changes:
Base.cfg:
set_shield_damage_multiplier - deprecated
base_shield_strength - float, stores the baseline damage reduction for any shielded POB. shield_reinforcement_threshold - float, damage that needs to be taken by the station to increase the damage reduction. shield_reinforcement_increment - float, amount by which shield reduction increases after passing each threshold.
base_vulnerability_window - [integer, integer], multiple entries allowed - a pair of integers specifying a time window where bases are vulnerable (in local timezone). For example: values of 12, 15 would mean the POBs are vulnerable from noon until 3PM

Example scenario:
Base has 1000 HP, baseline shield is 50%, reinforcement threshold is 100HP, and increment is 10%. Let's assume there are no repairs in the game. Until the base loses 100HP, it will take 50% reduced damage, then, until it loses another 100HP(down to 800) it will have a 60% reduction. Eventually, at 500HP the base will become invulnerable until the next vulnerability window begins. nullptr fix

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -356,6 +356,8 @@ void CoreModule::Spawn()
 			base->base_health = base->max_base_health;
 		pub::SpaceObj::SetRelativeHealth(space_obj, base->base_health / base->max_base_health);
 
+		base_shield_reinforcement_threshold = shield_reinforcement_threshold_flat + (shield_reinforcement_threshold_percent * base->max_base_health);
+
 		base->SyncReputationForBaseObject(space_obj);
 		if (set_plugin_debug > 1)
 			ConPrint(L"CoreModule::created space_obj=%u health=%f\n", space_obj, base->base_health);
@@ -600,8 +602,8 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 
 inline void CoreModule::AddDmgTakenToThresholdCounterAndReinforceShield(float dmgTaken) {
 	damage_taken_since_last_threshold += dmgTaken;
-	if (damage_taken_since_last_threshold >= shield_reinforcement_threshold) {
-		damage_taken_since_last_threshold -= shield_reinforcement_threshold;
+	if (damage_taken_since_last_threshold >= base_shield_reinforcement_threshold) {
+		damage_taken_since_last_threshold -= base_shield_reinforcement_threshold;
 		shield_strength_multiplier += shield_reinforcement_increment;
 	}
 }
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -12,7 +12,8 @@
 #include "Main.h"
 #include <hookext_exports.h>
 
-CoreModule::CoreModule(PlayerBase *the_base) : Module(TYPE_CORE), base(the_base), space_obj(0), dont_eat(false), dont_rust(false)
+CoreModule::CoreModule(PlayerBase *the_base) : Module(TYPE_CORE), base(the_base), space_obj(0), dont_eat(false), 
+dont_rust(false), shield_strength_multiplier(base_shield_strength), damage_taken_since_last_threshold(0)
 {
 }
 
@@ -583,13 +584,11 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 		return curr_hitpoints;
 	}
 
-	if (base->invulnerable == 0)
+	if (base->shield_state != PlayerBase::SHIELD_STATE_OFFLINE && shield_strength_multiplier < 1.0 && true == globalBaseVulnerabilityStatus && base->invulnerable == 0)
 	{
-		// Reduce the damage to 10% if the shield is or will be online.
-		if (base->shield_state != PlayerBase::SHIELD_STATE_OFFLINE)
-		{
-			return curr_hitpoints - ((curr_hitpoints - new_hitpoints) * set_shield_damage_multiplier);
-		}
+		float damageTaken = ((curr_hitpoints - new_hitpoints) * (1 - shield_strength_multiplier));
+		AddDmgTakenToThresholdCounterAndReinforceShield(damageTaken);
+		return curr_hitpoints - damageTaken;
 	}
 	else
 	{
@@ -599,6 +598,13 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 	return 0.0f;
 }
 
+inline void CoreModule::AddDmgTakenToThresholdCounterAndReinforceShield(float dmgTaken) {
+	damage_taken_since_last_threshold += dmgTaken;
+	if (damage_taken_since_last_threshold >= shield_reinforcement_threshold) {
+		damage_taken_since_last_threshold -= shield_reinforcement_threshold;
+		shield_strength_multiplier += shield_reinforcement_increment;
+	}
+}
 
 bool CoreModule::SpaceObjDestroyed(uint space_obj)
 {
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -582,7 +582,7 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 		return curr_hitpoints;
 	}
 
-	if (base->shield_state != PlayerBase::SHIELD_STATE_OFFLINE && shield_strength_multiplier < 1.0 && true == globalBaseVulnerabilityStatus && base->invulnerable == 0)
+	if (base->shield_state != PlayerBase::SHIELD_STATE_OFFLINE && shield_strength_multiplier < 1.0 && !isGlobalBaseInvulnerabilityActive && base->invulnerable == 0)
 	{
 		float damageTaken = ((curr_hitpoints - new_hitpoints) * (1 - shield_strength_multiplier));
 		AddDmgTakenToThresholdCounterAndReinforceShield(damageTaken);
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -356,7 +356,10 @@ void CoreModule::Spawn()
 			base->base_health = base->max_base_health;
 		pub::SpaceObj::SetRelativeHealth(space_obj, base->base_health / base->max_base_health);
 
-		base_shield_reinforcement_threshold = shield_reinforcement_threshold_flat + (shield_reinforcement_threshold_percent * base->max_base_health);
+		if (shield_reinforcement_threshold_map.count(base->base_level))
+			base_shield_reinforcement_threshold = shield_reinforcement_threshold_map[base->base_level];
+		else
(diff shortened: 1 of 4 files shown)


Plugins/Public/base_plugin/Main.cpp
```diff
@@ -467,9 +478,28 @@ void LoadSettingsActual()
...
+					else if (ini.is_value("shield_reinforcement_increment"))
 					{
-						set_shield_damage_multiplier = ini.get_value_float(0);
+						shield_reinforcement_increment = ini.get_value_float(0);
+					}
+					else if (ini.is_value("base_shield_strength"))
+					{
+						base_shield_strength = ini.get_value_float(0);
+					}
+					else if (ini.is_value("base_vulnerability_window"))
```
dsyalex, line 497: I think only allowing hours here is a bit limiting.

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2757,3 +2788,33 @@ EXPORT PLUGIN_INFO* Get_PluginInfo()
...
+
+	if (baseVulnerabilityWindows.empty()) {
+		return false;
+	}
+
+	time_t tNow = time(0);
+	struct tm *t = localtime(&tNow);
+	uint currHour = t->tm_hour;
+	for (list<BASE_VULNERABILITY_WINDOW>::iterator i = baseVulnerabilityWindows.begin(); i != baseVulnerabilityWindows.end(); ++i){
+		if((i->start < i->end
```
dsyalex, line 2808: Let's leave a comment about what this very messy if statement trying to do. I think the main distinction this code is looking for is whether a vulnerability window is defined as e.g.:
* more usually e.g. 0100-0700 - in which cases bases are vulneable when the time is on/after 0100 AND before 0700
* slightly less obvious problem case e.g. 2100-0300, i.e. over a day boundary - in which case bases are vulnerable when the time is on/after 2100 OR before 0300.

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -90,8 +90,19 @@ uint set_tick_time = 16;
...
-/// If the shield is up then damage to the base is changed by this multiplier.
-float set_shield_damage_multiplier = 0.03f;
+
+// set of configurable variables defining the diminishing returns on damage during POB siege
+// POB starts at base_shield_strength, then every 'threshold' of damage taken, 
+// shield goes up in absorption by the 'increment'
+float shield_reinforcement_threshold_flat = 1000000.0f;
+float shield_reinforcement_threshold_percent = 0.0f;
+float shield_reinforcement_increment = 0.0f;
+float base_shield_strength = 0.97f;
```
dsyalex, line 100: why does this default to 0.97f instead of 1.0f? feels like something odd is going on here

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -90,8 +90,19 @@ uint set_tick_time = 16;
 // How much damage do we heal per repair cycle?
 uint repair_per_repair_cycle = 60000;
 
-/// If the shield is up then damage to the base is changed by this multiplier.
-float set_shield_damage_multiplier = 0.03f;
+
+// set of configurable variables defining the diminishing returns on damage during POB siege
+// POB starts at base_shield_strength, then every 'threshold' of damage taken, 
+// shield goes up in absorption by the 'increment'
```
dsyalex, line 96: does not explain 'flat' vs. 'percent' distinction

dsyalex (changes requested):
I don't understand the logic behind modifying shield damage behaviour as damage is taken, but I have more obvious issues with the code right now

Plugins/Public/base_plugin/Main.cpp
Aingar: 'base shield strength' is the starting damage reduction. I used the base 97%, along with 0.0f increments so that in case of no configuration applied at all, the siege mechanics will continue to work as it does now on live, which is static 97% the whole day. Values are overwritten by the config files, if present.

Plugins/Public/base_plugin/Main.cpp
Aingar: Wanted an ability to customize the curve of POB toughness to core level.
I suppose I could just make a map where we strictly define threshold size from a map<coreLvl, threshold>?

Plugins/Public/base_plugin/Main.cpp
Aingar: Added the comment

Plugins/Public/base_plugin/Main.cpp
Aingar: It is, but then we'd add another layer of complexity to the formula, after discussing with Haste, it was considered 'good enough'

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -595,6 +596,13 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 	return 0.0f;
 }
 
+inline void CoreModule::AddDmgTakenToThresholdCounterAndReinforceShield(float dmgTaken) {
```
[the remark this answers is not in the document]
Aingar: done

Plugins/Public/base_plugin/Main.cpp
Aingar: Update: I have implemented the map of corelvl->shield reinforcement threshold.

Plugins/Public/base_plugin/Main.h
```diff
@@ -144,6 +153,7 @@ class CoreModule : public Module
 	float SpaceObjDamaged(uint space_obj, uint attacking_space_obj, float curr_hitpoints, float new_hitpoints);
 	bool SpaceObjDestroyed(uint space_obj);
 	void SetReputation(int player_rep, float attitude);
+	void AddDmgTakenToThresholdCounterAndReinforceShield(float dmgTaken);
```
dsyalex, line 156: no longer present?

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2757,3 +2784,36 @@ EXPORT PLUGIN_INFO* Get_PluginInfo()
...
+	struct tm *t = localtime(&tNow);
+	uint currHour = t->tm_hour;
+	// iterate over configured vulnerability periods to check if we're in one.
+	// - in case of timeStart < timeEnd, eg. 5-10, the base will be vulnerable between 5AM and 10AM.
+	// - in case of timeStart > timeEnd, eg. 23-2, the base will be vulnerable after 11PM and before 2AM, so a period of 3 hours.
+	for (list<BASE_VULNERABILITY_WINDOW>::iterator i = baseVulnerabilityWindows.begin(); i != baseVulnerabilityWindows.end(); ++i){
+		if((i->start < i->end 
+			&& i->start <= currHour && i->end > currHour)
+		|| (i->start > i->end
+			&& (i->start <= currHour || i->end > currHour))) {
```
dsyalex, line 2810: should we add a condition here for `i->start == i->end` ? otherwise I think a bad config could make future base vulnerability windows not count - they would be vulnerable all the time as this code would fail and it'd fall through to the default return

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2757,3 +2784,36 @@ EXPORT PLUGIN_INFO* Get_PluginInfo()
...
+	if (baseVulnerabilityWindows.empty()) {
+		return false;
+	}
+
+	time_t tNow = time(0);
+	struct tm *t = localtime(&tNow);
+	uint currHour = t->tm_hour;
+	// iterate over configured vulnerability periods to check if we're in one.
+	// - in case of timeStart < timeEnd, eg. 5-10, the base will be vulnerable between 5AM and 10AM.
+	// - in case of timeStart > timeEnd, eg. 23-2, the base will be vulnerable after 11PM and before 2AM, so a period of 3 hours.
```
dsyalex, line 2805: we don't really need to say "a period of 3 hours" here, the key point I was getting at was one situation uses an OR and one uses an AND and it's deliberate and thought out

Plugins/Public/base_plugin/Main.cpp
dsyalex: ... I don't really feel like you've explained anything here but I suppose I should consider it separately with new code

Plugins/Public/base_plugin/Main.cpp
dsyalex: Aingar clarified this over Discord - the 0.97f magic number is not entirely new, it's just that there was previously a `set_shield_damage_multiplier = 0.03f` now replaced with a definition of `0.97f` and `set_shield_damage_multiplier` references changed to `1 - shield_strength_multiplier`.

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2757,3 +2784,36 @@ EXPORT PLUGIN_INFO* Get_PluginInfo()
 	p_PI->lstHooks.push_back(PLUGIN_HOOKINFO((FARPROC*)&Plugin_Communication_CallBack, PLUGIN_Plugin_Communication, 11));
 	return p_PI;
 }
+
+void ResetShieldStrength() {
```
dsyalex, line 2788: Should this be named ResetAllBasesShieldStrength or something?

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -768,3 +768,16 @@ float PlayerBase::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 
 	return 0.0f;
 }
+
+// Reset shield to default strength
+void PlayerBase::ResetShieldStrength() {
+	CoreModule *mod;
```
dsyalex, line 774: Shouldn't this just be a part of line 776 where you set `mod`? This feels like the old JavaScript variable hoisting we used to do back in the day.

dsyalex (commented):
Some minor things to look at, but this is making sense to me now so this PR is on track to go in soon.

Plugins/Public/base_plugin/Main.cpp
Aingar: If config would contain a `i->start == i->end` entry, it'd skip the `baseVulnerabilityWindows.empty()` check, and fail the highlighted check, resulting in all bases being permanently invulnerable.
The only way to make bases permanently vulnerable would be to either configure no vulnerability entries, or have entry/entries spanning the entire 24 hour range.

Plugins/Public/base_plugin/Main.cpp
Aingar: Good catch, updated to avoid naming overlap.

Plugins/Public/base_plugin/Main.cpp
dsyalex: good point, not sure what I was thinking when I wrote this

dsyalex (approved):

dsyalex:
We're gonna try this out

PR #167, opened 2022-12-10
Added progressive Wear and Tear

Aingar:
New config file properties:
Base.cfg:

damage_multiplier - float, float, float - FromHP, ToHP, Modifier - if base hp is between FromHP and ToHP (percent values from 0.0 to 1.0), multiply the wear and tear damage applied by Modifier value. Multiple entries allowed for different ranges, but only the first modifier matching a particular hitpoint percentage is ever applied. Logic defaults Modifier to 1 if no entry found.

no_crew_damage_multiplier - float - If station has insufficient crew to maintain the POB(Core level x200), the base, multiply the wear and tear by this value.

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -456,37 +456,33 @@ void CoreModule::RepairDamage(float max_base_health)
 
 bool CoreModule::Timer(uint time)
 {
-	if (space_obj && set_holiday_mode)
-	{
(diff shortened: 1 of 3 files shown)


dsyalex (commented):
Looks good and almost ready to approve, but we should be careful to point out that only the first modifier matching a hitpoint range is ever applied.

dsyalex:
I meant in the PR description really.

dsyalex:
Turns out I can edit it! All done here

PR #168, opened 2022-12-10
Added Gold Ore Refinery FactoryModule, major rewrite of FactoryModule.

Aingar:
Ore Refinery added, help menu only specifies Gold Ore but recipes can be added to .cfg without any additional code changes.

FactoryModule rewrite changes:
-Ability to select an item by name or order number, 'infotext' and 'shortcut_number' properites in recipe .cfg  file respectively.
-Ability to add recipes to .cfg file without having to touch code -Ability to set a recipe to run continuously via 'loop_production' property -Ability to specify quantity of produced goods via 'produced_amount' property -Ability to specify the valid factory type for a given recipe via 'factory_type' property, which should match the 'nickname' property on the module recipe -Ability to start and stop construction of equipment without having to specify module index (Refinery only at the moment) -Lowered the amount of data written to the POB .ini files, as most of redundant recipe info is pulled from the recipe .cfg file instead of duplicated for every instance of a FactoryModule.

adjusted .cfg recipe templates:
base_recipe_items.cfg
[recipe]
nickname - unchanged 
produced_item = unchanged
factory_type = string, has to match a module recipe nickname to allow crafting 
recipe_type = string, 'module_recipe','factory_recipe','refinery_recipe', used to distinguish between types 
shortcut_number = number, replaces old hardcoded numbers for crafting 
produced_amount = number, how many items are crafted upon success 
loop_production = boolean, if true, recipe loops upon completion infotext = unchanged
cooking_rate = unchanged
reqlevel = unchanged
consumed = unchanged

base_recipe_modules.cfg
[recipe]
nickname = module_unchanged
produced_item = unchanged
infotext = unchanged
shortcut_number = same as above
cooking_rate = unchanged
recipe_type = same as above
reqlevel = unchanged
consumed = unchanged

Planned in near future: 
-Dynamically generated help menu for Factory and Refinery modules (so new recipes are added without having to recompile) -Applying the index-less navigation to non-Refinery Factory Modules


Aingar:
The TODO's from original commit message are now included.

On top of them I made the Factory modules respect owner-specified cargo limits. For example, recipe producing 100 Gold units will stall if POB is specified to accept up to 1000 units, and 950 are present on the station.

Next in line is split of /shop price command into several, more convenient commands, accepting string input on top of numbers, but I'll await criticism for current changes before adding more stuff.

PR #170, opened 2022-12-23
POB QoL changes

Aingar:
bugfix; /bank deposit will no longer work (and straight up delete credits) outside of POBs

new command: /base supplies
prints crew, FOW and repair material counts.
obraz


dsyalex (changes requested):
See Discord - this should be multiple separate PRs

PR #172, opened 2022-12-23
new '/base supplies' command conveniently listing crew, FOW and repairs

Aingar:


Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1825,6 +1828,52 @@ namespace PlayerCommands
...
+			PrintUserCmdText(client, L"WARNING, CREW COUNT TOO LOW");
+		}
+		PrintUserCmdText(client, L"Crew: %u onboard", base->HasMarketItem(set_base_crew_type));
+
+		PrintUserCmdText(client, L"Crew supplies:");
+		for (map<uint, uint>::iterator i = set_base_crew_consumption_items.begin(); i != set_base_crew_consumption_items.end(); ++i) {
+			const GoodInfo *gi = GoodList::find_by_id(i->first);
+			if (gi)
+			{
+				PrintUserCmdText(client, L"|    %s: %u", HkGetWStringFromIDS(gi->iIDSName).c_str(), base->HasMarketItem(i->first));
```
dsyalex, line 1856: Why do I feel like there was some horrible nasty gotcha with HkGetWStringFromIDS or two at some point... Never mind, should be fine.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1825,6 +1828,52 @@ namespace PlayerCommands
...
+			}
+		}
+		
+		uint foodCount = 0;
+		for (map<uint, uint>::iterator i = set_base_crew_food_items.begin(); i != set_base_crew_food_items.end(); ++i) {
+			foodCount += base->HasMarketItem(i->first);
```
dsyalex, line 1862: So for supplies and repair items we do per-item breakdowns, but food is summed up? That just seems inconsistent?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1825,6 +1828,52 @@ namespace PlayerCommands
...
+			return;
+		}
+
+		if (!clients[client].admin && !clients[client].viewshop)
+		{
+			PrintUserCmdText(client, L"ERROR: Access denied");
+			return;
+		}
+
+		if (base->HasMarketItem(set_base_crew_type) < base->base_level * 200) {
```
dsyalex, line 1846: I would be slightly happier with:
```
uint baseItemCount = base->HasMarketItem(set_base_crew_type);
uint baseItemNeed = base->base_level * 200;
if (baseItemCount < baseItemNeed) {
```
That way we would avoid multiple operations like this (relying on the compiler to get the order of operations right instead of being unambiguous, though I suppose with these types it couldn't work any other way), and we could also reuse baseItemCount below too.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1825,6 +1828,52 @@ namespace PlayerCommands
...
+		PlayerBase *base = GetPlayerBaseForClient(client);
+		if (!base)
+		{
+			PrintUserCmdText(client, L"ERR Not in player base");
+			return;
+		}
+
+		if (!clients[client].admin && !clients[client].viewshop)
+		{
+			PrintUserCmdText(client, L"ERROR: Access denied");
```
dsyalex, line 1842: This should probably actually be `ERR Access denied` for consistency with most of the rest of the plugin, instead of /shop.

dsyalex (commented):
Nothing majorly wrong here from what I can tell, some minor things inline that I'd prefer addressed before merge.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: In case of supplies and repair items, base consumes them all at an equal rate whenever they are available (and hull needs repairing in case of repair items), which makes you want to keep them stocked in equal amounts. In case of food however, base simply consumes the first item it finds from a rather lengthy list of food items, which includes several duplicates of "Food Rations" item. Also, 95%+ of bases uses a singular food item, which is the one that happens to be sold at the nearest solar.

As such, simply producing a sum of all food items is more convenient to a vast majority, if not all base owners.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: Agreed, submitted.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: Submitted.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: This part I shamelessly copied from the ShowShopStatus which renders the /shop window. Fixing something that isn't broken is something I try to avoid :)

dsyalex: Oh I think there's something broken about it, but we'll keep the function call.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1825,6 +1828,54 @@ namespace PlayerCommands
...
+			return;
+		}
+
+		if (!clients[client].admin && !clients[client].viewshop)
+		{
+			PrintUserCmdText(client, L"ERR Access denied");
+			return;
+		}
+
+        uint crewItemCount = base->HasMarketItem(set_base_crew_type);
```
dsyalex, line 1846: Something weird about the indentation here and on the two lines below. Mix of spaces and tabs?

dsyalex (changes requested):

dsyalex (approved):

PR #174, opened 2022-12-27
Refinery rework atomized

Aingar:
EDIT: Can be closed, a more up-to-date version will be submitted to a patch-branch when made available.

Resubmitting #168  in more atomized form, where each commit is responsible for doing one thing, for ease of review.


dsyalex:
per discussion, to be reopened against a new patch branch

PR #177, opened 2023-01-30
Jettison All now adheres to NoTrade commodity list

Aingar:
While player-trade and individual item jettison is blocked by Alley's plugin Player Restriction feature, 'Jettison All' allows the player to bypass this block. This commit rectifies that.

### DIFF
--- a/Plugins/Public/alley/PlayerRestrictions.cpp
+++ b/Plugins/Public/alley/PlayerRestrictions.cpp
@@ -1095,7 +1095,6 @@ void JettisonCargo(unsigned int iClientID, struct XJettisonCargo const &jc)
 				{
 					returncode = SKIPPLUGINS_NOFUNCTIONCALL;
 					PrintUserCmdText(iClientID, L"ERR you can't jettison %s.", stows(i->second).c_str());
-					return;
 				}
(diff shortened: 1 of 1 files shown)


Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -675,6 +675,19 @@ bool UserCmd_JettisonAll(uint iClientID, const wstring &wscCmd, const wstring &w
 			pub::IsCommodity(item->iArchID, flag);
 			if (!item->bMounted && flag)
 			{
+				bool skipItem = false;
+				for (map<uint, string>::iterator i = notradelist.begin(); i != notradelist.end(); ++i)
+				{
+					if (i->first == item->iArchID)
+					{
+						PrintUserCmdText(iClientID, L"ERR you can't jettison %s.", stows(i->second).c_str());
```
dsyalex, line 683: I'm not sure this is technically something we should respond with ERR to. Just silently don't drop it if the player is not allowed?

dsyalex (changes requested):

dsyalex (approved):

PR #178, opened 2023-02-23
fix: BalanceMagic no longer stops other plugins

Aingar:
Instead of sending our own damage entry (losing a lot of metadata such as attacker info) and stopping other plugins on the same hook from executing, we just override the new HP value that will then be 'naturally' passed on to be processed by the server and other plugins "As Chris Roberts intended"

Done by sending the 'newHealth' float as reference instead of as a copy, akin to FLHook4.


dsyalex (approved):
nice

PR #179, opened 2023-03-13
Autobuy now repairs component, doesn't require FLServer running in legacy compatibility mode

Aingar:
Disclaimer: Other plugins depending on DynPackets will most likely still require the Win7 compatibility mode to work properly, unless we'd apply the ST6 solution to them as well.


dsyalex:
I've reviewed the differences between the st6.h provided here and https://github.com/TheStarport/FLHookSDK/blob/e54b414e4620eace2555c51a3903922da96b5f5d/include/FLCore/st6.h, and all changes to that file back prior to the renaming back to FriendlyFire's 2020 revisions. They seem ok

dsyalex (approved):

PR #183, opened 2023-03-20
probable fix for "CD doesn't always interrupt cloak" bug

Aingar:
I couldn't personally reproduce, could be caused by FLHook/plugin config?
0x15 is another cruise disruptor damage type, and is used in other places (such as HyperJump plugin) for comparisons


dsyalex:
If you can't reproduce the bug, how do you intend to test the fix?

dsyalex (approved):

PR #184, opened 2023-03-20
New: Shipclass based purchase restrictions

Aingar:
- added a basic check for shipclass immediately on item purchase
- said check extends to commodities as well as equipment
- old logic of checking full configuration on base launch remains unchanged

### DIFF
--- a/Plugins/Public/alley/PlayerRestrictions.cpp
+++ b/Plugins/Public/alley/PlayerRestrictions.cpp
@@ -1202,7 +1202,7 @@ void __stdcall ReqAddItem(unsigned int iArchID, char const *Hardpoint, int count
 	*/
 
 	returncode = DEFAULT_RETURNCODE;
-	if (reverseTrade[iClientID] == true)
+	if (reverseTrade[iClientID])
 	{
 		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
 	}
@@ -1211,7 +1211,7 @@ void __stdcall ReqAddItem(unsigned int iArchID, char const *Hardpoint, int count
 void __stdcall GFGoodBuy(struct SGFGoodBuyInfo const &gbi, unsigned int client)
 {
 	returncode = DEFAULT_RETURNCODE;
-	if (false == SCI::CanBuyItem(gbi.iGoodID, client)) {
+	if (!SCI::CanBuyItem(gbi.iGoodID, client)) {
 		const GoodInfo* gi = GoodList::find_by_id(gbi.iGoodID);
 		switch (gi->iType) {
 			case GOODINFO_TYPE_COMMODITY: {
@@ -1237,7 +1237,7 @@ void __stdcall GFGoodBuy(struct SGFGoodBuyInfo const &gbi, unsigned int client)
 void __stdcall ReqChangeCash(int cash, unsigned int iClientID)
 {
 	returncode = DEFAULT_RETURNCODE;
-	if (reverseTrade[iClientID] == true)
+	if (reverseTrade[iClientID])
 	{
 		reverseTrade[iClientID] = false;
 		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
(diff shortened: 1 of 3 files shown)


Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1199,8 +1200,61 @@ void __stdcall ReqAddItem(unsigned int iArchID, char const *Hardpoint, int count
...
+	if (reverseTrade[iClientID] == true)
+	{
+		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
+	}
+}
+
+void __stdcall GFGoodBuy(struct SGFGoodBuyInfo const &gbi, unsigned int client)
+{
+	returncode = DEFAULT_RETURNCODE;
+	if (false == SCI::CanBuyItem(gbi.iGoodID, client)) {
```
dsyalex, line 1214:
```suggestion
	if (!SCI::CanBuyItem(gbi.iGoodID, client)) {
```

Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1199,8 +1200,61 @@ void __stdcall ReqAddItem(unsigned int iArchID, char const *Hardpoint, int count
...
+		const GoodInfo* gi = GoodList::find_by_id(gbi.iGoodID);
+		switch (gi->iType) {
+			case GOODINFO_TYPE_COMMODITY: {
+				PrintUserCmdText(client, L"ERR Your ship can't load this cargo");
+				break;
+			}
+			case GOODINFO_TYPE_EQUIPMENT: {
+				PrintUserCmdText(client, L"ERR Your ship can't use this equipment");
```
dsyalex, line 1222: should we say 'use' or something more like 'store'? since this is more about purchasing rather than mounting

Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1199,8 +1200,61 @@ void __stdcall ReqAddItem(unsigned int iArchID, char const *Hardpoint, int count
...
+		return;
+	}
+
+	reverseTrade[client] = false;
+}
+
+void __stdcall ReqChangeCash(int cash, unsigned int iClientID)
+{
+	returncode = DEFAULT_RETURNCODE;
+	if (reverseTrade[iClientID] == true)
```
dsyalex, line 1240:
```suggestion
	if (reverseTrade[iClientID])
```

Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1199,8 +1200,61 @@ void __stdcall ReqAddItem(unsigned int iArchID, char const *Hardpoint, int count
...
+	{
+		reverseTrade[iClientID] = false;
+		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
+	}
 }
 
+void __stdcall ReqSetCash(int cash, unsigned int iClientID)
+{
+	returncode = DEFAULT_RETURNCODE;
+	if (reverseTrade[iClientID] == true)
```
dsyalex, line 1250:
```suggestion
	if (reverseTrade[iClientID])
```

Plugins/Public/alley/shipclass_item.cpp
```diff
@@ -279,6 +279,21 @@ void SCI::CheckItems(unsigned int iClientID)
 	return;
 }
 
+bool SCI::CanBuyItem(uint iArchID, uint iClientID) {
+
+	auto foundItem = shipclassitems.find(iArchID);
+	if (foundItem != shipclassitems.end()) {
+		Archetype::Ship* shipArch = Archetype::GetShip(Players[iClientID].iShipArchetype);
+		for (uint allowedClass : foundItem->second.canmount) {
```
dsyalex, line 287: just wondering if `canmount` is appropriate for purchasing restrictions
and on that subject don't we also need to be looking at `nomount`?

dsyalex (commented):
mostly minor comments inline, except the last one regarding canmount/nomount that I think warrants discussion

dsyalex (commented):
mostly minor comments inline, except the last one regarding canmount/nomount that I think warrants discussion

dsyalex:
Also can we re-title this something like "Introduce purchasing restrictions for commodities, based on per-shipclass equipment mount restriction system" ?

Plugins/Public/alley/PlayerRestrictions.cpp
Aingar: You can't buy equipment you don't have a free slot to immediately mount it into, so I find this one fitting.

PR #189, opened 2023-03-25
Siege rework enhancement

Aingar:
- moved shield variables from CoreModule to PlayerBase
- saving relevant variables to playerbase ini file (maintain the shield parameters in case of mid-siege crash)
- removed "holiday mode fixes all POBs" code block (again)
- set the shield to online immediately on server startup, preventing a free shot before the Timer() executes for the first time

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -576,15 +576,6 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 {
 	base->SpaceObjDamaged(space_obj, attacking_space_obj, curr_hitpoints, new_hitpoints);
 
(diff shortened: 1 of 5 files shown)


Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -576,23 +576,14 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 {
 	base->SpaceObjDamaged(space_obj, attacking_space_obj, curr_hitpoints, new_hitpoints);
 
-	if (set_holiday_mode || (base->basetype == "jumpgate") || (base->basetype == "jumphole") || (base->basetype == "airlock") || (base->basetype == "planet"))
```
dsyalex, line 579: don't we still need those basetype checks? or are we expecting to get away with it because they should have base->invulnerable = 1?

Aingar: Those basetypes should be always marked as logic = 0, invulnerable = 1, making those checks completely superfluous.

PR #191, opened 2023-03-28
Siege hotfix

CzarekMierzejewski:
- Fixed a case where base takes no damage from player fire if shield generator is not present or without fuel
- rewritten Main.cpp::HkCb_AddDmgEntry and CoreModule::SpaceObjDamaged methods for better readability.
- micro optimizations


dsyalex (commented):
this is looking like it might need a rebase prior to review

PR #198, opened 2023-05-08
Periodic CLoot object cleanup

Aingar:
Once every 7 minutes, run a cleanup routine:
Iterate over all CLoot objects (tractorable loot) objects in the game, grouped by the entity that dropped them.
Then if needed, discard all but the latest 25th per each spaceObjId.

Objective: Make it impossible to lag out the server by jettisoning an absurd amount of tiny ore lootables into space.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ _ReSharper*/
 
 *.dll
 *.pdb
+
+Plugins/Public/*/headers/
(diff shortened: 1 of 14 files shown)


Aingar:
Since it's heavily reliant on #208, I'd like to close this PR and resubmit under the patch branch instead.

dsyalex:
closed per request

PR #199, opened 2023-05-08
POB save cycle rework

Aingar:
Instead of saving all POBs once a minute, this change changes it to instead behave as a rolling update, where every second, one POB is saved until the iterator reaches the end of the list, then starts from the beginning, resulting in avoiding the massive load spike from opening and saving into ~200 files at once.

### DIFF
--- a/Plugins/Public/base_plugin/Main.cpp
+++ b/Plugins/Public/base_plugin/Main.cpp
@@ -28,6 +28,7 @@ map<uint, CLIENT_DATA> clients;
 
 // Bases
 map<uint, PlayerBase*> player_bases;
+map<uint, PlayerBase*>::iterator repairIterator = player_bases.begin();
 
 map<uint, bool> mapPOBShipPurchases;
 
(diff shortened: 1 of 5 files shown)


Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -87,14 +87,14 @@ bool PlayerBase::Timer(uint curr_time)
 				return true;
 		}
 	}
-
+	/*
```
dsyalex, line 90: delete the code please, we are using a version control system

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -829,6 +830,20 @@ void HkTimerCheckKick()
...
 		base->Timer(curr_time);
 	}
+	if (!player_bases.empty() && !set_holiday_mode) {
+		if (baseSaveIterator == player_bases.end()) {
+			baseSaveIterator = player_bases.begin();
+		}
+		bool saveSuccessful = false;
+		while (!saveSuccessful && baseSaveIterator != player_bases.end()) {
+			auto& pb = baseSaveIterator->second;
+			if (pb->logic == 1 || pb->invulnerable == 0) {
```
dsyalex, line 840: where does this criteria come from?

Aingar: We only need to open and save to a POB file if it has a chance to change. It means that it needs to either have passive logic enabled (passive damage, factories, healing etc) or it needs to be able to take damage.

Basically it excludes indestructible POBs that have no logic attached to them, such as nomad POB JHs in Omicron Iota.

dsyalex (approved):

PR #201, opened 2023-05-09
Conn: enabling /conn from POBs

Aingar:
In case of POB not being there, player will be sent to the proxy base in the POBs system.

### DIFF
--- a/Plugins/Public/conn_plugin/Main.cpp
+++ b/Plugins/Public/conn_plugin/Main.cpp
@@ -139,26 +139,24 @@ bool ValidateCargo(unsigned int client)
 	return true;
 }
 
-uint GetCustomBaseForClient(unsigned int client)
+void StoreReturnPointForClient(unsigned int client)
 {
-	// Pass to plugins incase this ship is docked at a custom base.
 	CUSTOM_BASE_IS_DOCKED_STRUCT info;
 	info.iClientID = client;
 	info.iDockedBaseID = 0;
 	Plugin_Communication(CUSTOM_BASE_IS_DOCKED, &info);
-	return info.iDockedBaseID;
-}
-
-void StoreReturnPointForClient(unsigned int client)
-{
 	// It's not docked at a custom base, check for a regular base
-	uint base = GetCustomBaseForClient(client);
-	if (!base)
+	if (info.iDockedBaseID)
+	{
+		HookExt::IniSetI(client, "conn.retbase", info.iDockedBaseID);
+		HookExt::IniSetI(client, "conn.retbasebackup", Players[client].iSystemID);
+	}
+	else
+	{
+		uint base;
 		pub::Player::GetBase(client, base);
-	if (!base)
-		return;
-
-	HookExt::IniSetI(client, "conn.retbase", base);
+		HookExt::IniSetI(client, "conn.retbase", base);
+	}
 }
 
 unsigned int ReadReturnPointForClient(unsigned int client)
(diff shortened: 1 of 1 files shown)


Plugins/Public/conn_plugin/Main.cpp
```diff
@@ -178,24 +176,27 @@ void MoveClient(unsigned int client, unsigned int targetBase)
 		return;
 
 	// No plugin handled it, do it ourselves.
-	unsigned int system;
-	pub::Player::GetSystem(client, system);
 	Universe::IBase* base = Universe::get_base(targetBase);
-
-	pub::Player::ForceLand(client, targetBase); // beam
-
-	// if not in the same system, emulate F1 charload
```
dsyalex, line 187: what happened to this, why do we no longer have this?

Aingar: I wondered what purpose this elaborate code that was emulating a F1 relog was, but when I got rid of it, I haven't noticed any changes, so I decided to get rid of 'excess fat'

dsyalex: I've done a spot of archaeology. It's old code taken from the depths of HkBeam.
It appeared some time since v1.5.5 in 2007, but pre-dates the TSP FLHook repository from March 2010.
If the mtime on src/flhook_src/source/HkFuncPlayers.cpp in https://forge.the-starport.net/attachments/download/10/flhook1.6.0_plugin.rar is to be believed it dates from 03/10/08 or earlier.
I've done some more digging and found https://the-starport.net/freelancer/forum/viewtopic.php?post_id=4425#forumpost4425 which might be the best explanation we're going to get
Having read that I'd prefer not to remove it

Aingar: Agreed

dsyalex (dismissed):
needs changing to keep the F1 emulation as it relates to 2008-era discovery+w0dk4 troubleshooting

Plugins/Public/conn_plugin/Main.cpp
```diff
@@ -177,25 +175,53 @@ void MoveClient(unsigned int client, unsigned int targetBase)
...
+		if (pub::GetBaseID(proxyBaseID, scProxyBase.c_str()) == -4)
+		{
+			PrintUserCmdText(client, L"Return impossible, contact staff");
+			return;
+		}
+		PrintUserCmdText(client, L"Player base renamed/destroyed, ship redirected to a proxy base");
+		pub::Player::ForceLand(client, proxyBaseID); // beam
+		if (base->iSystemID != system)
+		{
+			Server.BaseEnter(targetBase, client);
```
dsyalex, line 214: doesn't this stuff need to be proxyBaseID? but also, can't we move the ForceLand to after our `if (base) { ... } else { ... }` blocks and just do it once?

dsyalex (changes requested):

Plugins/Public/conn_plugin/Main.cpp
```diff
@@ -177,25 +187,39 @@ void MoveClient(unsigned int client, unsigned int targetBase)
...
+		string scProxyBase = (string)((const char*)sysInfo->nickname) + "_proxy_base";
+		uint proxyBaseID;
+		if (pub::GetBaseID(proxyBaseID, scProxyBase.c_str()) == -4)
+		{
+			PrintUserCmdText(client, L"Return impossible, contact staff");
+			return;
+		}
+		PrintUserCmdText(client, L"Player base renamed/destroyed, ship redirected to a proxy base");
+		pub::Player::ForceLand(client, proxyBaseID); // beam
+		if (base->iSystemID != system)
```
dsyalex, line 217: I'm not sure if this has actually been tested since I would expect trying to access `base->iSystemID` in this case is likely a null pointer access

dsyalex (changes requested):

dsyalex (approved):

PR #202, opened 2023-05-09
Cloak momentum based consumption

Aingar:
By default it replaces the cloak alert sound with cloak entry sound, to restore that, put the following into cloak.cfg

[General]
cloak_alert_sound = dsy_jumpdrive_survey

### DIFF
--- a/Plugins/Public/cloak_plugin/Cloak.cpp
+++ b/Plugins/Public/cloak_plugin/Cloak.cpp
@@ -49,6 +49,8 @@ struct CLOAK_ARCH
 	int iHoldSizeLimit;
 	map<uint, CLOAK_FUEL_USAGE> mapFuelToUsage;
 	bool bDropShieldsOnUncloak;
+	bool bBreakOnProximity;
+	float fRange;
 };
 
 struct CLOAK_INFO
@@ -103,6 +105,7 @@ static map<uint, CLOAK_ARCH> mapCloakingDevices;
 static map<uint, CDSTRUCT> mapCloakDisruptors;
 
 static set<uint> setJumpingClients;
+static uint CloakAlertSound = CreateID("cloak_osiris");
 
 void LoadSettings();
 
@@ -172,6 +175,13 @@ void LoadSettings()
 					{
 						device.bDropShieldsOnUncloak = ini.get_value_bool(0);
 					}
+					else if (ini.is_value("break_cloak_on_proximity"))
+					{
+						device.bBreakOnProximity = ini.get_value_bool(0);
+					}
+					else if (ini.is_value("detection_range")) {
+						device.fRange = ini.get_value_float(0);;
+					}
 				}
 				mapCloakingDevices[CreateID(device.scNickName.c_str())] = device;
 				++cloakamt;
@@ -208,6 +218,13 @@ void LoadSettings()
 				}
 				mapCloakDisruptors[CreateID(disruptor.nickname.c_str())] = disruptor;
 				++cdamt;
+			}
+			else if (ini.is_header("General")) {
+				while (ini.read_value()) {
+					if (ini.is_value("cloak_alert_sound")) {
+						CloakAlertSound = CreateID(ini.get_value_string());
+					}
+				}
 			}
 		}
 		ini.close();
@@ -475,7 +492,6 @@ void HkTimerCheckKick()
 				{
 					pub::SpaceObj::DrainShields(iShipID);
 
-
 					if ((curr_time % 5) == 0)
 					{
 						uint iShip;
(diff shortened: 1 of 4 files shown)


Plugins/Public/cloak_plugin/Cloak.cpp
```diff
@@ -100,6 +107,7 @@ static map<uint, CLIENTCDSTRUCT> mapClientsCD;
 static map<uint, CLOAK_ARCH> mapCloakingDevices;
 static map<uint, CDSTRUCT> mapCloakDisruptors;
 
+static uint CloakAlertSound = CreateID("cloak_osiris");
```
dsyalex, line 110: rename to lowerCamelCase please

Plugins/Public/cloak_plugin/Cloak.cpp
```diff
@@ -291,51 +309,40 @@ void SetState(uint iClientID, uint iShipID, int iNewState)
...
 	if (info.bAdmin)
 		return true;
 
-	if (!info.singleCloakConsumed && info.arch.singleUseCloak) {
-		return removeSingleFuel(iClientID, info);
-	}
-
-	else if (info.singleCloakConsumed)
-	{
+	if(setJumpingClients.find(iClientID) != setJumpingClients.end())
```
dsyalex, line 318: why are we adding logic around jumping while cloaking here? if you're jumping then you don't need any cloaking fuel removed?

Plugins/Public/cloak_plugin/Cloak.cpp
```diff
@@ -291,51 +309,40 @@ void SetState(uint iClientID, uint iShipID, int iNewState)
...
+			const auto& fuelUsage = info.arch.mapFuelToUsage[item->iArchID];
+			float currFuelUsage = fuelUsage.usageStatic;
+			if (fuelUsage.usageLinear != 0.0f || fuelUsage.usageSquare != 0.0f) {
+				Vector dir1;
+				Vector dir2;
+				pub::SpaceObj::GetMotion(iShipID, dir1, dir2);
+				float vecLength = sqrtf(dir1.x * dir1.x + dir1.y * dir1.y + dir1.z * dir1.z);
+				
+				currFuelUsage += fuelUsage.usageLinear * vecLength;
+				currFuelUsage += fuelUsage.usageSquare * vecLength * vecLength;
```
dsyalex, line 334: so wait... we want to add fuel usage based on speed... but also based on speed squared?

Plugins/Public/cloak_plugin/Cloak.cpp
```diff
@@ -291,51 +309,40 @@ void SetState(uint iClientID, uint iShipID, int iNewState)
...
+			}
+			uint totalFuelUsage = static_cast<uint>(max(currFuelUsage, 0.0f));
+			if (item->iCount >= totalFuelUsage)
 			{
-				pub::Player::RemoveCargo(iClientID, item->sID, fuel_usage);
+				pub::Player::RemoveCargo(iClientID, item->sID, totalFuelUsage);
 				return true;
 			}
+			if(info.arch.mapFuelToUsage.size() == 1)
```
dsyalex, line 342: is this some sort of weird optimisation for specific configs?

Plugins/Public/cloak_plugin/Cloak.cpp
Aingar: Should probably be renamed. When in a jump tunnel, player ship is moving at an insane speed, around 10kps. This was added not to instantly drain their fuel the moment they hit a jumphole/gate

Plugins/Public/cloak_plugin/Cloak.cpp
Aingar: Correct, we want the formula to scale exponentially with speed, requiring us to use a quadratic formula to achieve wanted consumption at specific speed breakpoints. It also enables the player to traverse a larger distance if they're willing to move at lower speed.

Plugins/Public/cloak_plugin/Cloak.cpp
Aingar: If we're using a single fuel commodity and we just finished processing it, there's no reason to continue looking, so we break the loop early.

Plugins/Public/cloak_plugin/Cloak.cpp
dsyalex: very well

Plugins/Public/cloak_plugin/Cloak.cpp
dsyalex: Oh that's interesting to hear written down and stated as fact. I always had a suspicion that while in the jump tunnel the player is actually still in one of the systems somewhere moving at speed, but never proved it. Makes sense.

Plugins/Public/cloak_plugin/Cloak.cpp
dsyalex: ok I guess, stops us from having to iterate through the rest of the player's equipDescList

dsyalex (changes requested):
variable renaming and then this is good to go

Plugins/Public/cloak_plugin/Cloak.cpp
```diff
@@ -962,6 +952,14 @@ int __cdecl Dock_Call(unsigned int const &iShip, unsigned int const &iDockTarget
 	return 0;
 }
 
+void __stdcall SystemSwitchOut(uint iClientID, FLPACKET_SYSTEM_SWITCH_OUT& switchOutPacket)
+{
+	// in case of SERVER_PACKET hooks, first argument is junk data before it gets processed by the server.
```
dsyalex, line 957: what does this mean? iClientID Is wrong and that's why we're going to `HkGetClientIDByShip(switchOutPacket.shipId)` instead??

dsyalex (changes requested):

dsyalex (approved):

PR #197, opened 2023-05-08
Siege hotfix

Aingar:

PR #204, opened 2023-05-13
Add dmg entry optimizations

Aingar:
Correcting signature for HkCb_AddDmgEntry to be in line with Core.
Rehooking PVEController to use ShipDestroyed instead, for notable performance improvement.

### DIFF
--- a/Plugins/Public/base_plugin/Main.cpp
+++ b/Plugins/Public/base_plugin/Main.cpp
@@ -22,6 +22,7 @@
 #include "Main.h"
 #include <sstream>
 #include <hookext_exports.h>
+#include <unordered_map>
 
 // Clients
 map<uint, CLIENT_DATA> clients;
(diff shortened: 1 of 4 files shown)


Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2013,15 +2014,15 @@ void __stdcall CShip_destroy(CShip* ship)
...
+	auto& i = spaceobj_modules.find(space_obj);
 	if (i != spaceobj_modules.end())
 	{
 		returncode = SKIPPLUGINS;
 		i->second->SpaceObjDestroyed(space_obj);
 	}
 }
 
-void __stdcall HkCb_AddDmgEntry(DamageList *dmg, unsigned short p1, float& damage, enum DamageEntry::SubObjFate fate)
+void __stdcall HkCb_AddDmgEntry(DamageList *dmg, unsigned short sID, float& newHealth, enum DamageEntry::SubObjFate& fate)
```
dsyalex, line 2025: think you've missed the part where you change the references from `p1` to `sID` and from `damage` to `newHealth`

dsyalex (changes requested):
checked base/cloak, not pvecontroller

Plugins/Public/base_plugin/Main.cpp
Aingar: Reverting the names back to normal.
Proper renaming of the parameters within the method happened in the #197

dsyalex (approved):

PR #206, opened 2023-05-15
Missiles: add configurable list of projectiles not triggering the alert

Aingar:
Uses: Flaks and 'dumbfire' missiles which are actually seekers purely due to better handling even without any actual tracking.


Aingar:
Outdated, all features already part of #226

PR #209, opened 2023-05-15
Permanent Mobile Docking 2: Electric Boogaloo

Aingar:
Permanent player docking is back.


Zentor88:
Discussed just now, using a dedicated proxy base rather than the current system's is a great set up. "In Transit" system sounds good. 

Would recommend undock cooldown to simulate ongoing repairs before immediately rejoining combat, as well as increased costs to both repair and buy regens.

PR #211, opened 2023-05-21
Jettison: Adjusting Alley notrade to allow partial jettison

Aingar:
Define a percentage of cargo that will be successfully jettisoned, rest will be lost. Default behavior for undefined percentage or a defined zero will be a standard jettison block.

Also made pulling names for error messages dynamic instead of relying on config.

### DIFF
--- a/Plugins/Public/alley/PlayerRestrictions.cpp
+++ b/Plugins/Public/alley/PlayerRestrictions.cpp
@@ -257,7 +257,7 @@ void LoadSettings()
 				{
 					if (ini.is_value("tr"))
 					{
-						notradelist[CreateID(ini.get_value_string(0))] = ini.get_value_float(1);
+						notradelist[CreateID(ini.get_value_string(0))] = min(1.0f, max( 0.0f, ini.get_value_float(1)));
 					}
 				}
 			}
@@ -1111,7 +1111,7 @@ void JettisonCargo(unsigned int iClientID, struct XJettisonCargo const &jc)
 			uint amountToJettison = static_cast<uint>(jc.iCount * noTradeGood->second);
 			PrintUserCmdText(iClientID, L"%u units of %ls jettisoned, %u units lost in the process.", jc.iCount, goodName.c_str(), jc.iCount - amountToJettison);
 			pub::Player::RemoveCargo(iClientID, static_cast<ushort>(jc.iSlot), jc.iCount);
-			if (amountToJettison) {
+			if (amountToJettison > 0) {
 				uint shipId;
 				uint sysId;
 				Vector pos;
--- a/Plugins/Public/alley/PlayerRestrictions.cpp
+++ b/Plugins/Public/alley/PlayerRestrictions.cpp
@@ -1119,6 +1119,7 @@ void JettisonCargo(unsigned int iClientID, struct XJettisonCargo const &jc)
 				pub::Player::GetSystem(iClientID, sysId);
 				pub::Player::GetShip(iClientID, shipId);
 				pub::SpaceObj::GetLocation(shipId, pos, ori);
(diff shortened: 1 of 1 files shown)


Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1082,42 +1086,62 @@ void __stdcall SystemSwitchOutComplete(unsigned int iShip, unsigned int iClientI
...
+		const GoodInfo* gi = GoodList::find_by_id(noTradeGood->first);
+		wstring goodName = HkGetWStringFromIDS(gi->iIDSName).c_str();
+		if (noTradeGood->second == 0.0f)
+		{
+			PrintUserCmdText(iClientID, L"ERR you can't jettison %ls.", goodName.c_str());
+		}
+		else
+		{
+			uint amountToJettison = static_cast<uint>(jc.iCount * noTradeGood->second);
+			PrintUserCmdText(iClientID, L"%u units of %ls jettisonned, %u units lost in the process.", jc.iCount, goodName.c_str(), jc.iCount - amountToJettison);
```
dsyalex, line 1112:
```suggestion
			PrintUserCmdText(iClientID, L"%u units of %ls jettisoned, %u units lost in the process.", jc.iCount, goodName.c_str(), jc.iCount - amountToJettison);
```

Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1082,42 +1086,62 @@ void __stdcall SystemSwitchOutComplete(unsigned int iShip, unsigned int iClientI
...
+		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
+		const GoodInfo* gi = GoodList::find_by_id(noTradeGood->first);
+		wstring goodName = HkGetWStringFromIDS(gi->iIDSName).c_str();
+		if (noTradeGood->second == 0.0f)
+		{
+			PrintUserCmdText(iClientID, L"ERR you can't jettison %ls.", goodName.c_str());
+		}
+		else
+		{
+			uint amountToJettison = static_cast<uint>(jc.iCount * noTradeGood->second);
```
dsyalex, line 1111: so the config is supposed to be a float between 0 and 1 - if they configure it higher, or negative, we could have some issues here, right?

Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -1082,42 +1086,62 @@ void __stdcall SystemSwitchOutComplete(unsigned int iShip, unsigned int iClientI
...
+			PrintUserCmdText(iClientID, L"%u units of %ls jettisonned, %u units lost in the process.", jc.iCount, goodName.c_str(), jc.iCount - amountToJettison);
+			pub::Player::RemoveCargo(iClientID, static_cast<ushort>(jc.iSlot), jc.iCount);
+			if (amountToJettison) {
```
dsyalex, line 1114: Let's just be explicit here:
```suggestion
			if (amountToJettison > 0) {
```

Plugins/Public/alley/PlayerRestrictions.cpp
Aingar: Correct. I know that for zero specifically it still jettisons a single unit of cargo. Above 1 would allow to generate cargo on the spot. Will clamp those values on config load to 0 and 1 respectively.

Plugins/Public/alley/PlayerRestrictions.cpp
Aingar: Agreed

dsyalex (changes requested):
clamp config values and then this is good to go

PR #213, opened 2023-05-23
EventUtilities: new commands adjusted to old convention.

Aingar:


dsyalex (approved):
we'll need to revisit ArgCharname later but this seems reasonable

PR #215, opened 2023-05-27
Implementing POB deployment range limits to surrounding solars.

Aingar:


Plugins/PluginUtilities/PluginUtilities.cpp
```diff
@@ -938,6 +938,100 @@ Quaternion HkMatrixToQuaternion(Matrix m)
...
+			continue;
+		for (int j = 0; j < 3; j++)
+		{
+			if (j == col)
+				continue;
+			minorMatrix[counter] = m.data[i][j];
+			counter++;
+		}
+	}
+	return minorMatrix[0] * minorMatrix[3] - minorMatrix[1] * minorMatrix[2];
```
dsyalex, line 965: Shouldn't either this function, or the callers, be multiplying the result by m[row][col]?

Plugins/PluginUtilities/PluginUtilities.cpp
```diff
@@ -938,6 +938,100 @@ Quaternion HkMatrixToQuaternion(Matrix m)
...
 
+Matrix TransposeMatrix(Matrix& m)
+{
+	return { {
+		{m.data[0][0], m.data[1][0], m.data[2][0]},
+		{m.data[0][1], m.data[1][1], m.data[2][1]},
+		{m.data[0][2], m.data[1][2], m.data[2][2]}} };
+}
+
+float MatrixDeterminant(Matrix& m, uint row, uint col)
```
dsyalex, line 949: This is actually like... MatrixSubDeterminant or something. It works on the 2x2 matrices that form a 3x3 matrix, ignoring the rows and columns identified by `row` and `col` and taking the remaining values as a 2x2 matrix

dsyalex (commented):
Comments for PluginUtilities. SDK looks good. Will check base plugin next.

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -45,6 +45,19 @@ set<uint> bannedSystemList;
...
+float minMiningDistance = 30000;
+float minPlanetDistance = 2500;
+float minStationDistance = 10000;
+float minLaneDistance = 5000;
+float minJumpDistance = 15000;
+float minDistanceMisc = 2500;
+
+/// Deployment command cooldown trackimg
+unordered_map<uint, uint> deploymentCooldownMap;
+uint deploymentCooldownDuration = 60;
```
dsyalex, line 59: we need a cooldown?

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -870,6 +915,13 @@ void HkTimerCheckKick()
 			ExportData::ToJSON();
 		}
 	}
+
+	for (auto& cooldown : deploymentCooldownMap)
+	{
+		cooldown.second--;
```
dsyalex, line 921: We have lots of RAM but we don't like doing extra stuff per-tick. This is probably small enough to be fine, but it's not a design I'd like to encourage. Instead of having to operate every tick or w/e, it might be better to store an expiry timestamp in memory, only clean it up when a client logs off.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				if (!zone->lootableZone)
+					continue;
+
+				if (minMiningDistance > 0 && cfield->near_field(pos))
+				{
+					PrintUserCmdText(client, L"You can't deploy inside a mining field!");
+					return false;
+				}
+				// pretend zone is actually centered 0,0,0 and unrotated
+				// Subtract zone positon from player position so their relative position is the same
```
dsyalex, line 1903:
```suggestion
				// Subtract zone position from player position so their relative position is the same
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				playerPos.z -= zone->vPos.z;
+				
+				//Then rotate player position by inverse of original zone position, to make the relative positions truly in sync
+
+				Matrix InvertedZoneRot = InverseMatrix(zone->mRot);
+				playerPos = VectorMatrixMultiply(playerPos, InvertedZoneRot);
+
+				//Now, player position is effectively a vector from center of mining zone to player position.
+				//We normalize the vector, then apply original zone rotation and stretch it by zone sizes
+				Vector V2 = NormalizeVector(playerPos);
```
dsyalex, line 1916:
```suggestion
				Vector v2 = NormalizeVector(playerPos);
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				Matrix EllipseSizeMat = { {{zone->vSize.x,0,0},{0,zone->vSize.y,0},{0,0,zone->vSize.z}} };
+				V2 = VectorMatrixMultiply(V2, EllipseSizeMat);
```
dsyalex, line 1918:
```suggestion
				v2 = VectorMatrixMultiply(v2, EllipseSizeMat);
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				V2 = VectorMatrixMultiply(V2, zone->mRot);
```
dsyalex, line 1919:
```suggestion
				v2 = VectorMatrixMultiply(v2, zone->mRot);
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				
+				//Now that we have our aproximated closest point to the mining zone, we move it back so comparison with player position yields proper result
+				V2.x += zone->vPos.x;
```
dsyalex, line 1922:
```suggestion
				v2.x += zone->vPos.x;
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				V2.y += zone->vPos.y;
```
dsyalex, line 1923:
```suggestion
				v2.y += zone->vPos.y;
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				V2.z += zone->vPos.z;
```
dsyalex, line 1924:
```suggestion
				v2.z += zone->vPos.z;
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+
+				float distance = HkDistance3D(V2, pos);
```
dsyalex, line 1926:
```suggestion
				float distance = HkDistance3D(v2, pos);
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				foundSystemMatch = true;
+			}
+
+			float distance = HkDistance3D(solar->get_position(), pos);
+			switch (solar->iType)
+			{
+				case OBJ_PLANET:
+				case OBJ_MOON:
+				{
+					if (distance < minPlanetDistance + solar->get_radius()) // In case of planets, we only care about distance from actual surface, since it can vary wildly
```
dsyalex, line 1967:
```suggestion
					if (distance < (minPlanetDistance + solar->get_radius())) // In case of planets, we only care about distance from actual surface, since it can vary wildly
```

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,151 @@ namespace PlayerCommands
...
+				V2.y += zone->vPos.y;
+				V2.z += zone->vPos.z;
+
+				float distance = HkDistance3D(V2, pos);
+
+				if (distance < minMiningDistance)
+				{
+					if (zone->idsName)
+					{
+						PrintUserCmdText(client, L"Distance to %ls too close, minimum distance: %um.", HkGetWStringFromIDS(zone->idsName).c_str(), static_cast<uint>(minMiningDistance));
```
dsyalex, line 1932: We seem to do this `static_cast<uint>(someFloat)` thing a lot - how safe is that? What if someone specifies a negative value in the config?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1979,6 +2124,49 @@ namespace PlayerCommands
...
 		}
+		//passed cargo check, now make the distance check
+
+		uint systemId;
+		Vector position;
+		Matrix rotation;
+		pub::Player::GetSystem(client, systemId);
+		pub::SpaceObj::GetLocation(ship, position, rotation);
+		Rotate180(rotation);
+		TranslateX(position, rotation, 1000);
```
dsyalex, line 2135: so this is doing what, pretending the player's position is 1K behind them?

dsyalex (changes requested):

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: 1k in front of her, it's the exact copy paste of calculating coordinates for base deployment from PlayerBase() constructor.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: For negative values, casting float into uint causes an underflow to occur, generating an effectively (for range finding purposes) infinite value. Could add a check at config load level and clamp the value to zero

Plugins/Public/base_plugin/Main.cpp
Aingar: Iteration on every solar and checking distance is not exactly a cheap operation, we don't want someone to be able to spam it and potentially lag out the server by inputting this several times a second via a macro.

Plugins/Public/base_plugin/Main.cpp
Aingar: Fair, will rewrite with this in mind.

Plugins/Public/base_plugin/Main.cpp
dsyalex: makes sense

Plugins/Public/base_plugin/PlayerCommands.cpp
dsyalex: makes sense

dsyalex (dismissed):
Need to resolve questions around MatrixDeterminant, cooldown technique, and clamping config values for static_cast<uint> safety.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,243 @@ namespace PlayerCommands
...
+	bool CheckSolarDistances(uint client, uint systemID, Vector pos)
+	{
+		// Other POB Check
+		for (const auto& base : player_bases)
+		{
+			if (base.second->basetype != "legacy" 
+				|| base.second->invulnerable 
+				|| !base.second->logic
+				|| base.second->system != systemID
+				|| (base.second->position.x == pos.x
```
dsyalex, line 1894: Why do we want to allow two bases at exactly the same spot?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1882,6 +1882,243 @@ namespace PlayerCommands
...
+			float distance = HkDistance3D(pos, base.second->position);
+			if (distance < minOtherPOBDistance)
+			{
+				if (client)
+				{
+					PrintUserCmdText(client, L"Player base %ls is too close! Minimum distance: %um", base.second->basename.c_str(), static_cast<uint>(minOtherPOBDistance));
+				}
+				else
+				{
+					ConPrint(L"Base is too close to another Player Base");
```
dsyalex, line 1908: Let's provide base.second->basename here too

dsyalex (changes requested):

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: It's a bit of a hacky way to detect the base being compared with itself without passing an extra parameter to the method, and skipping processing if that's the case since distance would be 0

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: Will do

Plugins/Public/base_plugin/PlayerCommands.cpp
dsyalex: lol ok

PR #220, opened 2023-05-31
TradeLane: Fix TL state desync bug when crossing systems. Add Trade Lane ban list for ship archs.

Aingar:
When you jump into a system, you always see tradelanes as alive, even if they were disabled by an evil pirate a moment earlier. Without the fix, you could dock and escape with pirate having no counterplay or ability to chase.

Also adds a configurable list of ships banned from using trade lanes. Ships banned from using trade lanes also cannot form onto other ships, but can be formed onto. This was done to prevent 'formation trade lane' behavior from used as a bypass for the block.


dsyalex (commented):
Leave a comment explaining the formation-blocking behaviour and update the PR description to discuss the lane-banning feature you're also adding here, then this can go in

dsyalex (approved):

PR #225, opened 2023-06-06
Docking: More organic Dock_Call overrides

Aingar:
IMPORTANT: REQUIRES DOCK_CALL METHOD ADJUSTMENTS ON PRIVATE PLUGINS!

Sadly we're unable to cleanly override RequestEvent hook at the moment, as such this will not have effect on neutral-no-dock restriction.


Zentor88:
The concept and bug fix for pob jumpholes is great, but def needs Alex to look at due to the apparent updates also needed on DSAce

PR #228, opened 2023-06-09
Jettison: Trigger animation on /jettisonall. Use appropriate loot crate model

Aingar:


Plugins/Public/alley/PlayerRestrictions.cpp
```diff
@@ -718,8 +719,23 @@ bool UserCmd_JettisonAll(uint iClientID, const wstring &wscCmd, const wstring &w
...
+
+				if (isFirstJettisonItem)
+				{
+					isFirstJettisonItem = false;
+					XJettisonCargo jettisonCargo;
+					jettisonCargo.iShip = iShip;
+					jettisonCargo.iCount = item->iCount;
+					jettisonCargo.iSlot = item->iID;
+					pub::Player::SendNNMessage(iClientID, pub::GetNicknameId("cargo_jettisoned"));
+					Server.JettisonCargo(iClientID, jettisonCargo);
```
dsyalex, line 731: is it implied here that this is performing HkRemoveCargo (or rather the equivalent thereof) as well behind the scenes?

Aingar: Yes, it performs the full jettison sequence, removal of cargo, jettisoning it into space, playing cargo hold animation and disabling shields temporarily. Only the 'cargo jettison' needs to be played manually

dsyalex (approved):

PR #231, opened 2023-06-25
Dock_Call bugfixes

Aingar:
Fixes the issue where without explicitly doing a Server.RequestCancel() on a docking request that was changed by a plugin from success to failure, causes the server to periodically (once every 90 seconds) to re-attempt the dock.

Also fixes the issue where ACCESS_DENIED status did not play an appropriate voiceline on its own.


dsyalex (changes requested):
As discussed in Discord there's something a little off to how the iCancel changes are being handles within PlayerCntl

dsyalex (approved):
We think the iCancel/dockPort behaviour is actually correct on closer inspection

PR #233, opened 2023-07-07
Proximity client-side detection

Aingar:
All proximity based processing for cloak disuption/ally alerting is now ran clientside in real-time.

Appropriate clienthook changes made on the private repository.


Aingar:
List of changes:

Cloak fuel consumption scales with speed using a quadratic formula,.making standing still consume little fuel, and moving causing progressively larger consumption.

Cloaks now get disrupted via proximity to nongrouped player ship.

Cloaks still produce sound when within disruption range of a grouped ship

Cloak distance check moved clientside and massively increase in frequency for improved responsiveness and server performance

dsyalex:
This is good to go once the commits are squashed and the rationale made clear, and the relevant private repo PR approved.

Aingar:
Moved to #248

PR #234, opened 2023-07-08
Events plugin update - Trade

Zentor88:
- Adding a Description Field that will be passed into the Info Popup dialog
- Adding optional Ship class requirements for Trade Events
- Adding a Popup Dialog that explains all the pertinent info: sell location, restrictions, bonus amount, description

### DIFF
--- a/Plugins/Public/event/Main.cpp
+++ b/Plugins/Public/event/Main.cpp
@@ -28,10 +28,9 @@
 #include <hookext_exports.h>
 #include "minijson_writer.hpp"
 #include <set>
-#include <unordered_map>
 
 static int set_iPluginDebug = 0;
-unordered_map <uint, wstring> shipclassnames;
+map <uint, wstring> shipclassnames;
 
 /// A return code to indicate to FLHook if we want the hook processing to continue.
 PLUGIN_RETURNCODE returncode;
@@ -48,11 +47,10 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
 	{
 		if (set_scCfgFile.length() > 0)
 			LoadSettings();
-		HkLoadStringDLLs();
 	}
 	else if (fdwReason == DLL_PROCESS_DETACH)
 	{
-		HkUnloadStringDLLs();
+		
 	}
 	return true;
 }
@@ -68,7 +66,7 @@ struct TRADE_EVENT {
 	uint uHashID;
 	string sEventName;
 	string sURL;
-	wstring eventDescription;
+	string eventDescription;
 	int iBonusCash;
 	bool isClassRestricted = false;
 	wstring allowedShipClass;
@@ -231,7 +229,7 @@ void LoadSettings()
 					}
 					else if (ini.is_value("description"))
 					{
-						te.eventDescription = stows((ini.get_value_string(0));
+						te.eventDescription = ini.get_value_string(0);
 					}
 					else if (ini.is_value("startbase"))
 					{
@@ -585,6 +583,7 @@ void LoadSettings()
 	ConPrint(L"EVENT DEBUG: Loaded %u event player data\n", iLoaded3);
 
 	LoadIDs();
+	HkLoadStringDLLs();
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -692,7 +691,7 @@ void TradeEventNotice(uint iClientID, map<string, TRADE_EVENT>::iterator iter)
 	const Universe::ISystem* sys = Universe::get_system(base->iSystemID);
 	const GoodInfo* gi = GoodList::find_by_id(iter->second.uCommodityID);
 	if (iter->second.eventDescription.empty())
-		iter->second.eventDescription = L"None";
+		iter->second.eventDescription = "None";
 
 	wstring pagetext =
 		L"<TRA bold=\"true\"/><TEXT>You've entered the " + stows(iter->second.sEventName) + L" Event</TEXT><TRA bold = \"false\"/><PARA /><PARA />"
@@ -701,13 +700,13 @@ void TradeEventNotice(uint iClientID, map<string, TRADE_EVENT>::iterator iter)
 		L"<TEXT>Transport " + HkGetWStringFromIDS(gi->iIDSName) + L" to " + HkGetWStringFromIDS(base->iBaseIDS) + L" in the " + HkGetWStringFromIDS(sys->strid_name) + " system to complete this mission.</TEXT><PARA /><PARA />"
 
 		L"<TRA bold=\"true\"/><TEXT>Reward Bonus:</TEXT><TRA bold=\"false\"/><PARA/>"
-		L"<TEXT>$" + to_wstring(iter->second.iBonusCash) + " per unit</TEXT><PARA /><PARA />"
+		L"<TEXT>$" + stows(to_string(iter->second.iBonusCash)) + " per unit</TEXT><PARA /><PARA />"
 
 		L"<TRA bold=\"true\"/><TEXT>Ship Class:</TEXT><TRA bold=\"false\"/><PARA/>"
-		L"<TEXT>Restricted to: " + iter->second.allowedShipClass + "</TEXT><PARA /><PARA />"
+		L"<TEXT>Restricted to: " + iter->second.allowedShipClass.c_str() + "</TEXT><PARA /><PARA />"
 
 		L"<TRA bold=\"true\"/><TEXT>Description:</TEXT><TRA bold=\"false\"/><PARA/>"
-		L"<TEXT>" + iter->second.eventDescription + L"</TEXT><PARA /><PARA />";
+		L"<TEXT>" + stows(iter->second.eventDescription) + L"</TEXT><PARA /><PARA />";
 
 	wchar_t titleBuf[4000];
 	_snwprintf(titleBuf, sizeof(titleBuf), L"Trade Event Information");
@@ -715,8 +714,8 @@ void TradeEventNotice(uint iClientID, map<string, TRADE_EVENT>::iterator iter)
 	wchar_t buf[4000];
 	_snwprintf(buf, sizeof(buf), L"<RDL><PUSH/>%ls<POP/></RDL>", pagetext.c_str());
 
-	HkChangeIDSString(iClientID, 267144, titleBuf);
-	HkChangeIDSString(iClientID, 267145, buf);
+	HkChangeIDSString(iClientID, 500000, titleBuf);
+	HkChangeIDSString(iClientID, 500001, buf);
 
 	FmtStr caption(0, 0);
 	caption.begin_mad_lib(500000);
@@ -779,7 +778,7 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const& gbi, unsigned int iC
 				{
 					//PrintUserCmdText(iClientID, L"Checking ship class requirements....");
 					Archetype::Ship* TheShipArch = Archetype::GetShip(Players[iClientID].iShipArchetype);
-					wstring& classname = shipclassnames.find(TheShipArch->iShipClass)->second;
+					wstring classname = shipclassnames.find(TheShipArch->iShipClass)->second;
 					
 					if (classname != i->second.allowedShipClass)
 					{
@@ -788,7 +787,7 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const& gbi, unsigned int iC
 					}
 				}
 				else {
-					i->second.allowedShipClass = L"No Restrictions";
+					i->second.allowedShipClass = stows("No Restrictions").c_str();
 				}
 
 				HookExt::IniSetB(iClientID, "event.enabled", true);
--- a/Plugins/Public/event/Main.cpp
+++ b/Plugins/Public/event/Main.cpp
@@ -28,9 +28,10 @@
 #include <hookext_exports.h>
 #include "minijson_writer.hpp"
 #include <set>
+#include <unordered_map>
 
 static int set_iPluginDebug = 0;
-map <uint, wstring> shipclassnames;
+unordered_map <uint, wstring> shipclassnames;
 
 /// A return code to indicate to FLHook if we want the hook processing to continue.
 PLUGIN_RETURNCODE returncode;
@@ -45,12 +46,13 @@ BOOL WINAPI DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved)
 	// calls load settings on FLHook startup and .rehash.
 	if (fdwReason == DLL_PROCESS_ATTACH)
 	{
+		HkLoadStringDLLs();
 		if (set_scCfgFile.length() > 0)
 			LoadSettings();
 	}
 	else if (fdwReason == DLL_PROCESS_DETACH)
 	{
-		
+		HkUnloadStringDLLs();
 	}
 	return true;
 }
@@ -66,7 +68,7 @@ struct TRADE_EVENT {
 	uint uHashID;
 	string sEventName;
 	string sURL;
-	string eventDescription;
+	wstring eventDescription;
 	int iBonusCash;
 	bool isClassRestricted = false;
 	wstring allowedShipClass;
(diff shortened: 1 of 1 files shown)


Plugins/Public/event/Main.cpp
```diff
@@ -30,6 +30,7 @@
 #include <set>
 
 static int set_iPluginDebug = 0;
+map <uint, wstring> shipclassnames;
```
Aingar, line 33: Replace with unordered_map please.
It will require an additional #include

Plugins/Public/event/Main.cpp
```diff
@@ -64,7 +66,10 @@ struct TRADE_EVENT {
 	uint uHashID;
 	string sEventName;
 	string sURL;
+	string eventDescription;
```
Aingar, line 69: Please turn into wstring, and call stows only on config load instead of every time it's read in line 709.
If possible, extend that to the other strings in the struct as well.

Plugins/Public/event/Main.cpp
```diff
@@ -556,16 +583,16 @@ void LoadSettings()
 	ConPrint(L"EVENT DEBUG: Loaded %u event player data\n", iLoaded3);
 
 	LoadIDs();
-
+	HkLoadStringDLLs();
```
Aingar, line 586: Please move into DLL_PROCESS_ATTACH section of DllMain(), and add HkUnloadStringDLLs(); to the DETACH section

Plugins/Public/event/Main.cpp
```diff
@@ -556,16 +583,16 @@ void LoadSettings()
...
 }
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
 //Functions
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-FILE *Logfile = fopen("./flhook_logs/event_log.log", "at");
```
Aingar, line 566: In the future, please try to avoid such formatting changes bleeding into the PR when the line is otherwise untouched.

Plugins/Public/event/Main.cpp
```diff
@@ -658,7 +685,50 @@ void __stdcall CharacterSelect_AFTER(struct CHARACTER_ID const & cId, unsigned i
...
+		L"<TRA bold=\"true\"/><TEXT>Description:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>" + stows(iter->second.eventDescription) + L"</TEXT><PARA /><PARA />";
+
+	wchar_t titleBuf[4000];
+	_snwprintf(titleBuf, sizeof(titleBuf), L"Trade Event Information");
+
+	wchar_t buf[4000];
+	_snwprintf(buf, sizeof(buf), L"<RDL><PUSH/>%ls<POP/></RDL>", pagetext.c_str());
+
+	HkChangeIDSString(iClientID, 500000, titleBuf);
```
Aingar, line 717: Those two IDSes are already in use. There is an infocard in place reserved for FLHook stuff already.
Replace with 267144 and 267145 please.

Plugins/Public/event/Main.cpp
```diff
@@ -658,7 +685,50 @@ void __stdcall CharacterSelect_AFTER(struct CHARACTER_ID const & cId, unsigned i
...
+	HkChangeIDSString(iClientID, 500001, buf);
+
+	FmtStr caption(0, 0);
+	caption.begin_mad_lib(500000);
```
Aingar, line 721: ditto

Plugins/Public/event/Main.cpp
```diff
@@ -658,7 +685,50 @@ void __stdcall CharacterSelect_AFTER(struct CHARACTER_ID const & cId, unsigned i
...
+	caption.end_mad_lib();
+
+	FmtStr message(0, 0);
+	message.begin_mad_lib(500001);
```
Aingar, line 725: ditto

Plugins/Public/event/Main.cpp
```diff
@@ -703,6 +773,23 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const &gbi, unsigned int iC
...
+				// Is the event limited by Ship class?
+				if (i->second.isClassRestricted)
+				{
+					//PrintUserCmdText(iClientID, L"Checking ship class requirements....");
+					Archetype::Ship* TheShipArch = Archetype::GetShip(Players[iClientID].iShipArchetype);
+					wstring classname = shipclassnames.find(TheShipArch->iShipClass)->second;
+					
+					if (classname != i->second.allowedShipClass)
+					{
+						PrintUserCmdText(iClientID, L"DEBUG: Invalid ship class[%s], this Evnet is for [%s]'s", classname.c_str(), i->second.allowedShipClass.c_str());
```
Aingar, line 785: Fix the typo: 'Evnet'

Plugins/Public/event/Main.cpp
```diff
@@ -703,6 +773,23 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const &gbi, unsigned int iC
 					}
 				}
 
+				// Is the event limited by Ship class?
+				if (i->second.isClassRestricted)
+				{
+					//PrintUserCmdText(iClientID, L"Checking ship class requirements....");
+					Archetype::Ship* TheShipArch = Archetype::GetShip(Players[iClientID].iShipArchetype);
+					wstring classname = shipclassnames.find(TheShipArch->iShipClass)->second;
```
Aingar, line 781: Whenever possible, avoid copying large objects such as wstrings, instead read them via reference:
wstring& classname = ...

Plugins/Public/event/Main.cpp
```diff
@@ -658,7 +685,50 @@ void __stdcall CharacterSelect_AFTER(struct CHARACTER_ID const & cId, unsigned i
...
+		L"<TRA bold=\"true\"/><TEXT>You've entered the " + stows(iter->second.sEventName) + L" Event</TEXT><TRA bold = \"false\"/><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Mission Objective:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>Transport " + HkGetWStringFromIDS(gi->iIDSName) + L" to " + HkGetWStringFromIDS(base->iBaseIDS) + L" in the " + HkGetWStringFromIDS(sys->strid_name) + " system to complete this mission.</TEXT><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Reward Bonus:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>$" + stows(to_string(iter->second.iBonusCash)) + " per unit</TEXT><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Ship Class:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>Restricted to: " + iter->second.allowedShipClass.c_str() + "</TEXT><PARA /><PARA />"
```
Aingar, line 706: no need for the .c_str() here

Plugins/Public/event/Main.cpp
```diff
@@ -658,7 +685,50 @@ void __stdcall CharacterSelect_AFTER(struct CHARACTER_ID const & cId, unsigned i
...
+		iter->second.eventDescription = "None";
+
+	wstring pagetext =
+		L"<TRA bold=\"true\"/><TEXT>You've entered the " + stows(iter->second.sEventName) + L" Event</TEXT><TRA bold = \"false\"/><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Mission Objective:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>Transport " + HkGetWStringFromIDS(gi->iIDSName) + L" to " + HkGetWStringFromIDS(base->iBaseIDS) + L" in the " + HkGetWStringFromIDS(sys->strid_name) + " system to complete this mission.</TEXT><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Reward Bonus:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>$" + stows(to_string(iter->second.iBonusCash)) + " per unit</TEXT><PARA /><PARA />"
```
Aingar, line 703: pretty sure you can just 'to_wstring()' here instead

Plugins/Public/event/Main.cpp
```diff
@@ -703,6 +773,23 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const &gbi, unsigned int iC
...
+					wstring classname = shipclassnames.find(TheShipArch->iShipClass)->second;
+					
+					if (classname != i->second.allowedShipClass)
+					{
+						PrintUserCmdText(iClientID, L"DEBUG: Invalid ship class[%s], this Evnet is for [%s]'s", classname.c_str(), i->second.allowedShipClass.c_str());
+						return;
+					}
+				}
+				else {
+					i->second.allowedShipClass = stows("No Restrictions").c_str();
```
Aingar, line 790: Unnecessary c_str() and instead of converting string into a literal, just load it directly as wstring:
... = L"No Restrictions";

Plugins/Public/event/Main.cpp
```diff
@@ -556,16 +583,16 @@ void LoadSettings()
 	ConPrint(L"EVENT DEBUG: Loaded %u event player data\n", iLoaded3);
 
 	LoadIDs();
```
Aingar, line 585: It might be a bit much, but I think it would be neat if you could have a validation of a trade event. Namely, calling the following within a try-catch block, loading it in if it succeeds and printing error information on the screen via ConPrint and/or AddLog()

```
Universe::IBase* base = Universe::get_base()
const Universe::ISystem* sys = Universe::get_system()
const GoodInfo* gi = GoodList::find_by_id()
```

Aingar (changes requested):
Lots of formatting changes bled through, and some mismanagement of string-wstring conversions.

Plugins/Public/event/Main.cpp
```diff
@@ -658,7 +686,50 @@ void __stdcall CharacterSelect_AFTER(struct CHARACTER_ID const & cId, unsigned i
...
+		iter->second.eventDescription = L"None";
+
+	wstring pagetext =
+		L"<TRA bold=\"true\"/><TEXT>You've entered the " + stows(iter->second.sEventName) + L" Event</TEXT><TRA bold = \"false\"/><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Mission Objective:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>Transport " + HkGetWStringFromIDS(gi->iIDSName) + L" to " + HkGetWStringFromIDS(base->iBaseIDS) + L" in the " + HkGetWStringFromIDS(sys->strid_name) + " system to complete this mission.</TEXT><PARA /><PARA />"
+
+		L"<TRA bold=\"true\"/><TEXT>Reward Bonus:</TEXT><TRA bold=\"false\"/><PARA/>"
+		L"<TEXT>$" + to_wstring((iter->second.iBonusCash)) + " per unit</TEXT><PARA /><PARA />"
```
Aingar, line 704: unnecessary double parentheses

Plugins/Public/event/Main.cpp
Aingar: still needs the fix :]

Aingar (changes requested):
Just two formatting/typo changes and it's good to go

Aingar (approved):
5th time's a charm!
Approved.

Plugins/Public/event/Main.cpp
```diff
@@ -703,6 +774,23 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const &gbi, unsigned int iC
 					}
 				}
 
+				// Is the event limited by Ship class?
+				if (i->second.isClassRestricted)
+				{
+					//PrintUserCmdText(iClientID, L"Checking ship class requirements....");
+					Archetype::Ship* TheShipArch = Archetype::GetShip(Players[iClientID].iShipArchetype);
+					wstring& classname = shipclassnames.find(TheShipArch->iShipClass)->second;
```
dsyalex, line 782: what if TheShipArch->iShipClass isn't found in shipclassnames? and shouldn't the comparison take place against ->first?

Plugins/Public/event/Main.cpp
```diff
@@ -703,6 +774,23 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const &gbi, unsigned int iC
...
+					
+					if (classname != i->second.allowedShipClass)
+					{
+						PrintUserCmdText(iClientID, L"DEBUG: Invalid ship class[%s], this Event is for [%s]'s", classname.c_str(), i->second.allowedShipClass.c_str());
```
dsyalex, line 786:
```suggestion
						PrintUserCmdText(iClientID, L"DEBUG: Invalid ship class %s, this event is for %ss", classname.c_str(), i->second.allowedShipClass.c_str());
```

Plugins/Public/event/Main.cpp
```diff
@@ -703,6 +774,23 @@ void __stdcall GFGoodBuy_AFTER(struct SGFGoodBuyInfo const &gbi, unsigned int iC
 					}
 				}
 
+				// Is the event limited by Ship class?
+				if (i->second.isClassRestricted)
+				{
+					//PrintUserCmdText(iClientID, L"Checking ship class requirements....");
```
dsyalex, line 780: don't leave commented out code lying around like this

Plugins/Public/event/Main.cpp
```diff
@@ -64,7 +68,10 @@ struct TRADE_EVENT {
 	uint uHashID;
 	string sEventName;
 	string sURL;
+	wstring eventDescription;
 	int iBonusCash;
+	bool isClassRestricted = false;
+	wstring allowedShipClass;
```
dsyalex, line 74: are we not likely to want to allow multiple ship classes?

dsyalex (changes requested):
Comments inline. Do we know if somebody in particular wants to use this feature?

Plugins/Public/event/Main.cpp
Aingar: Definitely. Personally I'd suggest making either a vector or unordered_set of allowed transport classes. If empty, allow event for all ship types. Otherwise check against the set and proceed only on match.

Plugins/Public/event/Main.cpp
Aingar: This print should be hidden behind a debug flag or removed, we don't want players to see this message.

Plugins/Public/event/Main.cpp
Aingar: Comparison happens against they key in find(), he's just inline extracting the value of ->second.

I agree that assuming a match will be found might cause issues in the future.

Do just the find, ensure it's not end(), and proceed appropriately.

Plugins/Public/event/Main.cpp
Zentor88: No we don't want multiple ship classes, otherwise it wouldn'be tailored profit-wise for that class

Plugins/Public/event/Main.cpp
Zentor88: We should notify players they will not be a part of the bonus event, but in a non-DEBUG method.

Plugins/Public/event/Main.cpp
Aingar: If a route is tailored to trains, there's no reason that I can see to stop smaller transports or liners from participating despite having lower CSU.

Zentor88:
> Comments inline. Do we know if somebody in particular wants to use this feature?

Yes Events Team wants to use this.

Plugins/Public/event/Main.cpp
Zentor88: > If a route is tailored to trains, there's no reason that I can see to stop smaller transports or liners from participating despite having lower CSU.

In that scenario, you don't even need to make a shipclass restricted run. Cuz running anything smaller will be pointless

Zentor88: Thesse really have two usages:
Cute RP focused thing (Freighter smuggling xyz out of a warzone for instance)
Giving actual profitability to specific underplayed ship classes. i.e. Liners, Freighters

PR #239, opened 2023-07-13
Siege weaponry dmg override

Aingar:
As long as base is shielded, only specific siege weapons are capable of dealing damage (during vulnerability window), unless no siege guns are defined.

Attacking with non-siege gun will still trigger a shield, however.

### DIFF
--- a/Plugins/Public/base_plugin/CoreModule.cpp
+++ b/Plugins/Public/base_plugin/CoreModule.cpp
@@ -586,8 +586,22 @@ float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, floa
 		// shield offline, return expected damage without modifications
 		return new_hitpoints;
 	}
-	
-	float damageTaken = ((curr_hitpoints - new_hitpoints) * (1 - base->shield_strength_multiplier));
+
+	float damageTaken;
+	if (!siegeWeaponryMap.empty())
+	{
+		const auto& siegeDamageIter = siegeWeaponryMap.find(iDmgMunitionID);
+		if(siegeDamageIter == siegeWeaponryMap.end())
+			return curr_hitpoints;
+		else
+		{
+			damageTaken = siegeDamageIter->second * (1 - base->shield_strength_multiplier);
+		}
+	}
+	else
+	{
+		damageTaken = ((curr_hitpoints - new_hitpoints) * (1 - base->shield_strength_multiplier));
+	}
 
 	base->damage_taken_since_last_threshold += damageTaken;
 	if (base->damage_taken_since_last_threshold >= base->base_shield_reinforcement_threshold) {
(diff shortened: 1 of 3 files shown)


Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -575,19 +575,37 @@ bool CoreModule::Timer(uint time)
 float CoreModule::SpaceObjDamaged(uint space_obj, uint attacking_space_obj, float curr_hitpoints, float new_hitpoints)
 {
 	base->SpaceObjDamaged(space_obj, attacking_space_obj, curr_hitpoints, new_hitpoints);
+	
+	if(base->shield_state == PlayerBase::SHIELD_STATE_OFFLINE){
```
Lazrius, line 579: Fix if formatting

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -575,19 +575,37 @@ bool CoreModule::Timer(uint time)
...
+		const auto& siegeDamageIter = siegeWeaponryMap.find(iDmgMunitionID);
+		if (siegeDamageIter == siegeWeaponryMap.end())
+		{
+			//Siege gun(s) defined, but this is not one of them, no damage dealt
+			return curr_hitpoints;
+		}
+		else
+		{
+			//Even with siege gun damage override, it still takes shield strength into the account
+			damageTaken = siegeDamageIter->second * (1 - base->shield_strength_multiplier);
```
Lazrius, line 602: This one should be moved to a constexpr to denote what it actually means. Also if these are both floats, using an int and not 1.0f is going to raise warnings.

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -575,19 +575,37 @@ bool CoreModule::Timer(uint time)
...
+	}
+	else
+	{
+		damageTaken = ((curr_hitpoints - new_hitpoints) * (1 - base->shield_strength_multiplier));
 	}
-	
-	float damageTaken = ((curr_hitpoints - new_hitpoints) * (1 - base->shield_strength_multiplier));
 
 	base->damage_taken_since_last_threshold += damageTaken;
 	if (base->damage_taken_since_last_threshold >= base->base_shield_reinforcement_threshold) {
```
Lazrius, line 611: Fix if formatting

Lazrius (changes requested):
As a note, with these changes being made to PoBs I would highly suggest bringing their health values down and make it so all weapons in the game just do less as a base. The fact that PoBs have such high health values makes everything more complicated. I suggest following the "Divide by 100 rule"

Aingar:
> As a note, with these changes being made to PoBs I would highly suggest bringing their health values down and make it so all weapons in the game just do less as a base. The fact that PoBs have such high health values makes everything more complicated. I suggest following the "Divide by 100 rule"

Given how with this change we can define the magnitude of damage numbers dealt to a POB, reducing their HP pools to something readable is basically the first thing on the to-do list.

Plugins/Public/base_plugin/CoreModule.cpp
Aingar: base->shield_strength_multiplier is variable, so it can't be condensed into a constant expression. Adjusting the number literals.

Lazrius (approved):

dsyalex (approved):
Per Laz's approval

PR #241, opened 2023-07-16
BaseDeploy: Using native, more accurate method for zone distance check

Aingar:


Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1908,43 +1908,48 @@ namespace PlayerCommands
...
+						ConPrint(L"Base is too close to another Player Base, distance %um, name %ls", static_cast<uint>(distance), base.second->basename.c_str());
+					}
+					return false;
 				}
-				return false;
 			}
 		}
 
 		// Mining Zone Check
 		CmnAsteroid::CAsteroidSystem* csys = CmnAsteroid::Find(systemID);
```
Lazrius, line 1941: I would suggest a better variable name than csys

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1908,43 +1908,48 @@ namespace PlayerCommands
...
-		if (csys)
+		if (csys && minMiningDistance > 0)
 		{
 			for (CmnAsteroid::CAsteroidField* cfield = csys->FindFirst(); cfield; cfield = csys->FindNext())
 			{
 				auto& zone = cfield->zone;
 				if (!zone->lootableZone)
```
Lazrius, line 1947: Fix if

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1908,43 +1908,48 @@ namespace PlayerCommands
...
+		if (minOtherPOBDistance > 0)
+		{
+			for (const auto& base : player_bases)
 			{
-				if (client)
-				{
-					PrintUserCmdText(client, L"Player base %ls is too close! Minimum distance: %um", base.second->basename.c_str(), static_cast<uint>(minOtherPOBDistance));
-				}
-				else
+				if (base.second->basetype != "legacy"
```
Lazrius, line 1915: When you have an if like this you should document the control flow in a comment. Ideally avoid checks like this because this is very hard to understand. Especially with brackets being used near the end with things like position.

Lazrius: And while I know that bases were legacy class, it's worth documenting what that actually means to anyone who is going to look at this later.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1908,43 +1908,48 @@ namespace PlayerCommands
...
+					|| base.second->invulnerable
+					|| !base.second->logic
+					|| base.second->system != systemID
+					|| (base.second->position.x == pos.x
+						&& base.second->position.y == pos.y
+						&& base.second->position.z == pos.z))
+					continue;
```
Lazrius, line 1922: Fix if braces

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1908,43 +1908,48 @@ namespace PlayerCommands
...
+
+				float distance = HkDistance3D(pos, base.second->position);
+				if (distance < minOtherPOBDistance)
 				{
-					ConPrint(L"Base is too close to another Player Base, distance %um, name %ls", static_cast<uint>(distance), base.second->basename.c_str());
+					if (client)
+					{
+						PrintUserCmdText(client, L"Player base %ls is too close! Minimum distance: %um", base.second->basename.c_str(), static_cast<uint>(minOtherPOBDistance));
```
Lazrius, line 1929: Probably worth printing their current distance. There can be a misalignment of what the game's client/camera sees and what the server sees. Also you should format these numbers with commas and whatnot so they are readable. Reading that you are 3434231 to close would be very confusing.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1908,43 +1908,48 @@ namespace PlayerCommands
...
+					}
+					else
+					{
+						ConPrint(L"Base is too close to another Player Base, distance %um, name %ls", static_cast<uint>(distance), base.second->basename.c_str());
```
Lazrius, line 1933: You should log this to a file as well.

Lazrius (changes requested):
Does this handle jumpholes and other dockables?

Aingar:
> Does this handle jumpholes and other dockables?

Handles all CSolar objects, dockables, undockables, Trade Lanes, wrecks.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: Good point. Also made me realize the check as it was would allow players to plonk POBs next to admin placed objects.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: Nah, those client-less results are printed to console for admin use upon entering a command. They will just be copied and processed directly from there.

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: 5 digits are plenty readable imo. Adding current distances tho.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -2105,7 +2107,7 @@ namespace PlayerCommands
 						if (!idsName) idsName = solar->get_archetype()->iIdsName;
```
Lazrius, line 2107: Fix if

Lazrius (approved):

dsyalex (approved):
on the basis of Laz's approval

PR #242, opened 2023-07-16
NPC: Group fleets for individual control. Add goto command.

Aingar:
Pending testing.

### DIFF
--- a/Plugins/Public/npc_control/Main.cpp
+++ b/Plugins/Public/npc_control/Main.cpp
@@ -16,8 +16,8 @@
 #include <string>
 #include <time.h>
 #include <math.h>
-#include <list>
 #include <map>
+#include <unordered_set>
 #include <algorithm>
 #include <FLHook.h>
 #include <plugin.h>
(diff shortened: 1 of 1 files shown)


Plugins/Public/npc_control/Main.cpp
```diff
@@ -727,86 +864,71 @@ void AdminCmd_AIFleet(CCmds* cmds, wstring FleetName)
...
-};
+	RIGHT_CHECK(RIGHT_AICONTROL)
 
-USERCMD UserCmds[] =
-{
-	{ L"/autobuy", UserCmd_AutoBuy, L"Usage: /autobuy" },
-	{ L"/autobuy*", UserCmd_AutoBuy, L"Usage: /autobuy" },
-};
-*/
+	if (coordList.count(coordName))
```
treewyrm, line 906: Perhaps it's better to overwrite existing waypoint with new coords.

dsyalex (changes requested):
Still listed as "Pending testing"

PR #245, opened 2023-07-28
BaseDeploy: Add configurable credit cost

Aingar:
Made it so spamming a bunch of POBs is deterred by massive upfront credit cost.

### DIFF
--- a/Plugins/Public/base_plugin/Main.cpp
+++ b/Plugins/Public/base_plugin/Main.cpp
@@ -62,6 +62,9 @@ uint deploymentCooldownDuration = 60;
 /// Map of good to quantity for items required by construction ship
 map<uint, uint> construction_items;
 
+/// Construction cost in credits
+int construction_credit_cost = 0;
+
 /// list of items and quantity used to repair 10000 units of damage
 list<REPAIR_ITEM> set_base_repair_items;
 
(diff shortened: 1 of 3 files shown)


Zentor88:
Why not just control this by the cost of the commodity needed to deploy the base rather than additional code?

Aingar:
We'd need to either increase the price of construction materials (which may be just regular tradable/POB repair/POB crafting commodities), which would turn economy out of whack, or require a specific 'Modular Base Core' commodity sold arbitrarily at select few locations.

There's also a potential issue of 'what if someone stocked up on a bunch of those and rendered themselves immune to our price increases?'

It's much more foolproof to just siphon cash directly at the time of deployment.

Aingar:
> Why not just control this by the cost of the commodity needed to deploy the base rather than additional code?

We'd need to either increase the price of construction materials (which may be just regular tradable/POB repair/POB crafting commodities), which would turn economy out of whack, or require a specific 'Modular Base Core' commodity sold arbitrarily at select few locations.

There's also a potential issue of 'what if someone stocked up on a bunch of those and rendered themselves immune to our price increases?'

It's much more foolproof to just siphon cash directly at the time of deployment.

dsyalex (approved):

PR #248, opened 2023-08-14
Cloak rework: fuel, detection, clientside checks

Aingar:
One-commit version of #233

List of changes:

Cloak fuel consumption scales with speed using a quadratic formula,.making standing still consume little fuel, and moving causing progressively larger consumption.

Cloaks now get disrupted via proximity to nongrouped player ship.

Cloaks still produce sound when within disruption range of a grouped ship

Cloak distance check moved clientside and massively increased in frequency for improved responsiveness and server performance


dsyalex (approved):
cf #233

PR #229, opened 2023-06-16
Permanent Mobile Docking 2: Electric Boogaloo

Aingar:
Had to recreate the branch due to unintended commits creeping in.

PR #250, opened 2023-08-15
Permanent Mobile Docking - master branch PR

Aingar:
Reopening #229 directly to master (the other PR suddenly balooned in size after a merge conflict.)

### DIFF
--- a/Plugins/Public/mobiledocking_plugin/Main.cpp
+++ b/Plugins/Public/mobiledocking_plugin/Main.cpp
@@ -413,7 +413,7 @@ void HkTimerCheckKick()
 		if (V1mov.x > 5 || V1mov.y > 5 || V1mov.z > 5)
 		{
 			auto dockingName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(dd.dockingID));
-			PrintUserCmdText(dd.dockingID, L"Docking aborted due to %ls's movement", dockingName);
+			PrintUserCmdText(dd.dockingID, L"Docking aborted due to your movement");
 			PrintUserCmdText(dd.carrierID, L"Docking aborted due to %ls's movement", dockingName);
 			dockdata = dockingInProgress.erase(dockdata);
 			continue;
@@ -440,7 +440,7 @@ void HkTimerCheckKick()
 			if (distance > maxDockingDistanceTolerance)
 			{
 				auto dockingName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(dd.dockingID));
-				PrintUserCmdText(dd.dockingID, L"Docking aborted due to %ls's movement", dockingName);
+				PrintUserCmdText(dd.dockingID, L"Docking aborted due to your movement");
 				PrintUserCmdText(dd.carrierID, L"Docking aborted due to %ls's movement", dockingName);
 				dockdata = dockingInProgress.erase(dockdata);
 				continue;


Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2875,6 +2927,46 @@ void Plugin_Communication_CallBack(PLUGIN_MESSAGE msg, void* data)
...
+			{
+				info->lastBaseName = L"Destroyed Player Base";
+			}
+		}
+		else
+		{
+			info->lastBaseName = L"Object Unknown";
+		}
+	}
+	else if (msg == CUSTOM_BASE_GET_NAME)
```
dsyalex, line 2950: Something is fishy here - why is the code in here a duplicate of the above? I would say put it in an OR and just do it once, but I doubt it makes sense to do that

dsyalex (changes requested):

Source/FLHookPluginSDK/headers/plugin.h
```diff
@@ -250,12 +250,16 @@ enum PLUGIN_MESSAGE
 	DSACE_SPEED_EXCEPTION = 41,
 	CUSTOM_BASE_BEAM = 42,
 	CUSTOM_BASE_IS_DOCKED = 43,
-	CUSTOM_BASE_LAST_DOCKED = 46,
 	CLIENT_CLOAK_INFO = 44,
 	COMBAT_DAMAGE_OVERRIDE = 45,
+	CUSTOM_BASE_LAST_DOCKED = 46,
 	CUSTOM_JUMP = 47,
-	CUSTOM_IS_IT_POB = 50,
```
dsyalex, line 257: why are we re-numbering this?

Aingar: Fair, re-adjusting

Source/FLHookPluginSDK/headers/plugin.h
```diff
@@ -250,12 +250,15 @@ enum PLUGIN_MESSAGE
...
 	CUSTOM_REVERSE_TRANSACTION = 48,
+	CUSTOM_JUMP_CALLOUT = 49,
```
dsyalex, line 258: Nothing seems to actually receive these messages?

Source/FLHookPluginSDK/headers/plugin.h
```diff
@@ -365,11 +368,37 @@ struct CUSTOM_JUMP_STRUCT
 	uint iSystemID;
 };
 
+struct CUSTOM_JUMP_CALLOUT_STRUCT
+{
+	uint iClientID;
+	uint iSystemID;
+	Vector pos;
+	Matrix ori;
+	uint jumpType = 0;
```
dsyalex, line 377: Unused?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+		std::lock_guard<std::mutex> saveLock(saveMutex);
+		time_t curTime = time(0);
+		while (ini.read_header())
+		{
+			if (ini.is_header("CarrierData"))
+			{
+				CARRIERINFO ci;
+				wstring carrierName;
+				boolean doLoad = true;
+				while (ini.read_value() && doLoad)
```
dsyalex, line 154: don't we want to do this while ini.read_value() regardless of doLoad? Since we'll want a complete list of docked ships for the MoveOfflineShipToLastDockedSolar caller below

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+					}
 				}
 			}
 		}
 		ini.close();
 	}
 
+	if (ini.open(scCarrierDataFile.c_str(), false))
+	{
+		std::lock_guard<std::mutex> saveLock(saveMutex);
```
dsyalex, line 145: unused?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
 
-	// Normalize the docking modules available, with the number of people currently docked
-	mobiledockClients[iClientID].iDockingModulesAvailable = (mobiledockClients[iClientID].iDockingModulesInstalled - mobiledockClients[iClientID].mapDockedShips.size());
+		char datapath[MAX_PATH];
+		GetUserDataPath(datapath);
+		string path = string(datapath) + R"(\Accts\MultiPlayer\docking_module\mobile_docking.ini)";
+		FILE* file = fopen(path.c_str(), "w");
+		if (file)
+		{
+			std::lock_guard<std::mutex> saveLock(saveMutex);
```
dsyalex, line 233: unused?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
 
-// Temporary storage for client data to be handled in LaunchPosHook.
-CLIENT_DATA undockingShip;
+wstring GetLastBaseName(uint client)
+{
+	const auto& dockedInfo = idToDockedInfoMap[client];
+	const auto& baseInfo = Universe::get_base(dockedInfo->lastDockedSolar);
+	const auto& sysInfo = Universe::get_system(baseInfo->iSystemID);
+	wstring baseName;
+	wstring& sysName = HkGetWStringFromIDS(sysInfo->strid_name);
```
dsyalex, line 266: I'm slightly nervous about the ability for these HkGetWStringFromIDS calls to blow up in our faces if a mod developer screws up referential integrity (which they do all the time) and names a base after some non-existent infocard

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+
+	//remove the docked ship entry from carrier's data
+	if (nameToCarrierInfoMap.count(carrierName))
+	{
+		auto& dockedList = nameToCarrierInfoMap[carrierName].dockedShipList;
+		for (auto& iter = dockedList.begin(); iter != dockedList.end(); iter++)
+		{
+			if (*iter == dockedShipName)
+			{
+				std::lock_guard<std::mutex> saveLock(saveMutex);
```
dsyalex, line 332: unused?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
-					PrintUserCmdText(client, L"No proxy base in system detected. Contact a developer about this please. Required base: %ws", scProxyBase);
-					return;
-				}
+		}
 
-				mobiledockClients[client].carrierSystem = -1;
-			}
+		if (!shipAlreadyDocked)
+		{
+			PrintUserCmdText(dockingID, L"Carrier has no free docking capacity");
```
dsyalex, line 371: Should we send an access denied message or something at this point?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
-		}
+	//In case this ship was launched from another mobiledock, unregister it first.
+	RemoveShipFromLists(dockingName, false);
 
-		//Get the carrier ship information
-		uint carrierShip;
-		pub::Player::GetShip(carrier_client, carrierShip);
+	if (!shipAlreadyDocked)
+	{
+		std::lock_guard<std::mutex> saveLock(saveMutex);
```
dsyalex, line 384: unused again?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+		idToDockedInfoMap[dockingID] = &nameToDockedInfoMap[dockingName];
+		idToCarrierInfoMap[carrierID]->lastCarrierLogin = time(nullptr);
+		mobiledockClients[carrierID].iDockingModulesAvailable--;
+	}
 
-		uint clientShip;
-		pub::Player::GetShip(client, clientShip);
+	// Land the ship on the designated base
+	HkBeamById(dockingID, mobileDockingProxyBase);
+	PrintUserCmdText(carrierID, L"Ship docked");
```
dsyalex, line 397: Shouldn't we tell the carrier which ship docked?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+		{
+			dockdata = dockingInProgress.erase(dockdata);
+			continue;
+		}
+		Vector V1mov, V1rot;
+		pub::SpaceObj::GetMotion(dockingShipID, V1mov, V1rot);
+		if (V1mov.x > 5 || V1mov.y > 5 || V1mov.z > 5)
 		{
+			PrintUserCmdText(dd.dockingID, L"Docking aborted due to craft movement");
+			PrintUserCmdText(dd.carrierID, L"Docking aborted due to craft movement");
```
dsyalex, line 419: name the player?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+	returncode = DEFAULT_RETURNCODE;
+
+	// DELAYEDDOCK vector iteration
+	for (auto& dockdata = dockingInProgress.begin(); dockdata != dockingInProgress.end();)
+	{
+		auto& dd = dockdata->second;
+		const auto& dockingShipID = Players[dd.dockingID].iShipID;
+		if (dockingShipID == 0)
+		{
+			dockdata = dockingInProgress.erase(dockdata);
```
dsyalex, line 411: do we want to notify the carrier in this scenario?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+			continue;
+		}
 
-			if (iBaseID)
+		dd.timeLeft--;
+		if (!dd.timeLeft)
+		{
+			Vector pos;
+			Matrix _;
```
dsyalex, line 439: let's give this a better variable name

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+			float distance = HkDistance3D(pos, dd.startPosition);
+			if (distance > maxDockingDistanceTolerance)
 			{
-				// Set the flags which the PlayerLaunch_AFTER uses to handle teleporting to the carriers base
-				mobiledockClients[client].undockBase = Universe::get_base(iBaseID);
-				mobiledockClients[carrier_client].mapDockedShips.erase(clientName);
-				mobiledockClients[client].baseUndock = true;
-				return;
+				PrintUserCmdText(dd.dockingID, L"Docking aborted due to craft movement");
+				PrintUserCmdText(dd.carrierID, L"Docking aborted due to craft movement");
```
dsyalex, line 445: name the craft

Source/FLHookPluginSDK/headers/plugin.h
Aingar: The code for that was developed as part of the HyperJump rework. It won't work without it, but I guarantee it's there.
Otherwise we'd need a lot of duplicated code across branches that'd need to be synchronized.

Source/FLHookPluginSDK/headers/plugin.h
Aingar: It is used in the PlayerController versions of this callout. Again, wanted to avoid code duplication across branches.

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Excellent point, adjusting.

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Added

Plugins/Public/mobiledocking_plugin/Main.cpp
Lazrius: These are threading locks. They are scoped so they call mutex lock and then when they go out of scope they call unlock. Removing these would cause the server to crash with no indication of why.

Plugins/Public/mobiledocking_plugin/Main.cpp
Lazrius: See above

Plugins/Public/mobiledocking_plugin/Main.cpp
Lazrius: See above

Plugins/Public/mobiledocking_plugin/Main.cpp
Lazrius: See below

Plugins/Public/mobiledocking_plugin/Main.cpp
Lazrius: See below

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Given how the whole operation is between two people, you'd know if you moved (because you pressed something) or if the other one moved (because you didn't)

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Means the ship is no longer in space next to it, the only scenario I see it happening in would be docking on the station next to it, which would be communicated separately.

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Given how the whole operation is between two people, you'd know if you moved (because you pressed something) or if the other one moved (because you didn't)

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Renamed to 'dummy'

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: Added

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: HkGetWStringFromIDS() already defaults in case of a nonexistent IDS (I believe to "Object Unknown"?)

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -460,97 +792,100 @@ int __cdecl Dock_Call(unsigned int const &iShip, unsigned int const &iBaseID, in
...
-		status += L"<TEXT>" + XMLText(mobiledockClients[client].wscDockedWithCharname) + L"</TEXT><PARA/><PARA/>";
+		status += L"<TEXT>Currently docked on: " + XMLText(idToDockedInfoMap[client]->carrierName);
+		if (carrierID != -1)
+		{
+			// if carrier online, add system info.
+			const auto& sysInfo = Universe::get_system(Players[carrierID].iSystemID);
+			wstring& systemName = HkGetWStringFromIDS(sysInfo->strid_name);
+			status += L", " + systemName;
+		}
+		status += L"</TEXT><PARA/><PARA/>";
```
dsyalex, line 824:
```suggestion
		status += L"</TEXT>";

```

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -460,97 +792,100 @@ int __cdecl Dock_Call(unsigned int const &iShip, unsigned int const &iBaseID, in
...
-					mobiledockClients[iDockedClientID].carrierSystem = cship->iSystem;
-
-					JettisonShip(client, iDockedClientID);
-				}
-
-				// Clear the carrier from the list
-				mobiledockClients[client].mapDockedShips.clear();
-				mobiledockClients[client].iDockingModulesAvailable = mobiledockClients[client].iDockingModulesInstalled;
-			}
+			PrintUserCmdText(carrierID, L"%s has left your garrison, reclaiming hangar space.", charName);
```
dsyalex, line 833: garrison?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+}
+
+void StartDockingProcedure(uint dockingID, uint carrierID)
+{
 
-		// If carrier and docked ship are in same system.
-		if (Players[client].iSystemID == Players[carrier_client].iSystemID)
+	if (dockingPeriod)
+	{
+		uint shipId = Players[dockingID].iShipID;
```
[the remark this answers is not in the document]
Aingar: Thought about it and realised there's one very unlikely scenario, added handling.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -460,97 +792,100 @@ int __cdecl Dock_Call(unsigned int const &iShip, unsigned int const &iBaseID, in
...
 			}
+			counter++;
+		}
+		if (mobiledockClients[client].iDockingModulesAvailable)
+		{
+			PrintUserCmdText(client, L"Remaining free capacity: %u", mobiledockClients[client].iDockingModulesAvailable);
 		}
 		return true;
 	}
 	else if (wscCmd.find(L"/conn") == 0 || wscCmd.find(L"/return") == 0)
```
dsyalex, line 885: kinda feel like this should be done with inter-plugin communication rather than trying to process the user's commands first but I recognise it's pre-existing code

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -559,42 +894,51 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
+			if (charNumber <= dockedShipList.size())
+			{
+				selectedShip = dockedShipList.at(charNumber - 1);
+			}
+			else
+			{
+				PrintUserCmdText(client, L"Invalid ship index selected");
+				return true;
+			}
 		}
```
dsyalex, line 923: what if it's <= 0?

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,700 @@ void LoadSettings()
...
+	if (mobiledockClients[carrierShipID].iDockingModulesAvailable > 0)
+	{
+		return true;
 	}
-	return true;
+	if (!idToCarrierInfoMap.count(carrierShipID))
+	{
+		return false;
+	}
+	wstring dockingName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(dockingID));
```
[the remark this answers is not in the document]
Aingar: So the way this plugin works is that once the fighter player docks on the carrier, it's considered their 'respawn point', and they take up the spot until they're forcibly jettisoned by the carrier, or the fighter docks on another station.
As such, they must be capable of docking on a carrier if they're already 'registered' to it, since they have a reserved spot.

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -628,270 +972,182 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
+	{
+		idToDockedInfoMap[iClientID] = &nameToDockedInfoMap[charname];
+		uint carrierClientID = HkGetClientIdFromCharname(idToDockedInfoMap[iClientID]->carrierName.c_str());
+		if (carrierClientID != -1)
 		{
-			uint clientID;
-			try
+			const auto& shipInfo = Archetype::GetShip(Players[iClientID].iShipArchetype);
+			wstring& shipName = HkGetWStringFromIDS(shipInfo->iIdsName);
+			PrintUserCmdText(carrierClientID, L"INFO: Docked %ls, %ls is standing by for launch", shipName.c_str(), charname);
```
dsyalex, line 1075: Instead of saying docked this should talk about the docked ship having come online

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -628,270 +972,182 @@ bool UserCmd_Process(uint client, const wstring &wscCmd)
...
-			SERVERlines.push_back(ID + " " + Charname + " " + wstos(wscIP) + " " + State);
+			PrintUserCmdText(iClientID, L"Carrier currently offline. Last docked base: %ls", lastBaseName);
 		}
+	}
+}
+
+void __stdcall UseItemRequest(SSPUseItem const& p1, unsigned int iClientID)
+{
+	returncode = DEFAULT_RETURNCODE;
+	if (disableRegensOnDockAttempt && !dockingInProgress.count(iClientID))
```
dsyalex, line 1111: is this ! supposed to be here?

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: It does?

Plugins/Public/mobiledocking_plugin/Main.cpp
Aingar: That's the whole reason I hooked UseItemRequest?
If it's configured to do so, any ship currently undergoing the docking procedure has it's shields disabled, cannot use bots and batts, and moving aborts the sequence. It is to make 'bot/battery tanking' not an option during the process, making it much more high-risk in combat.

This comes in tandem with an official rule change explicitly stating that docking on a Docking Module does NOT constituted a pvp death, and facilities inside allow for a full refill (bots, batts, ammo + repairs)

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: Right but it's a many to one relationship, right? Each docking ship has exactly one carrier they can be docking to at any given time, so they don't need any disambiguation, but the carrier has many potential craft that could be docking to it at any one time.

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: see above

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: I think this is written not just confusingly but wrong - right now, if you set disableRegensOnDockAttempt to false, this code is going to block all regen usage regardless of docking status. What you actually want is:
```c++
void __stdcall UseItemRequest(SSPUseItem const& p1, unsigned int iClientID)
{
    if (disableRegensOnDockAttempt && dockingInProgress.count(iClientID))
    {
        static uint rejectSound = CreateID("ui_select_reject");
        pub::Audio::PlaySoundEffect(iClientID, rejectSound);
        return;
    }
}
```

dsyalex (changes requested):

Plugins/Public/mobiledocking_plugin/Main.cpp
dsyalex: Per discussion - it's L"" and the BHkLoadStringDLLs call was added to mobiledocking's DllMain

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,707 @@ void LoadSettings()
...
+		{
+			dockdata = dockingInProgress.erase(dockdata);
+			continue;
+		}
+		Vector V1mov, V1rot;
+		pub::SpaceObj::GetMotion(dockingShipID, V1mov, V1rot);
+		if (V1mov.x > 5 || V1mov.y > 5 || V1mov.z > 5)
 		{
+			auto dockingName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(dd.dockingID));
+			PrintUserCmdText(dd.dockingID, L"Docking aborted due to %ls's movement", dockingName);
```
dsyalex, line 416: this isn't quite what I had in mind - only the carrier needs the dockingName sent to it

Plugins/Public/mobiledocking_plugin/Main.cpp
```diff
@@ -192,266 +89,707 @@ void LoadSettings()
...
+			float distance = HkDistance3D(pos, dd.startPosition);
+			if (distance > maxDockingDistanceTolerance)
 			{
-				// Set the flags which the PlayerLaunch_AFTER uses to handle teleporting to the carriers base
-				mobiledockClients[client].undockBase = Universe::get_base(iBaseID);
-				mobiledockClients[carrier_client].mapDockedShips.erase(clientName);
-				mobiledockClients[client].baseUndock = true;
-				return;
+				auto dockingName = reinterpret_cast<const wchar_t*>(Players.GetActiveCharacterName(dd.dockingID));
+				PrintUserCmdText(dd.dockingID, L"Docking aborted due to %ls's movement", dockingName);
```
dsyalex, line 443: ditto

dsyalex (approved):

PR #251, opened 2023-08-22
Destroy CLoot on collision

Aingar:
CLoot containers are effectively unmovable objects, causing any ship that tries to fly into one to either quickly bounce off in case of smaller ships, or spin violently in case of caps. This destroys the CLoot containers when collided at above-trivial speed (default 25, configurable)

### DIFF
--- a/Plugins/Public/playercntl_plugin/Main.cpp
+++ b/Plugins/Public/playercntl_plugin/Main.cpp
@@ -588,8 +588,18 @@ namespace HkIServerImpl
 		returncode = DEFAULT_RETURNCODE;
 
 		// If spin protection is off, do nothing.
-		if (set_fSpinProtectMass == -1.0f)
+		if (!ci.dwTargetShip || set_fSpinProtectMass == -1.0f)
 			return;
+
+		uint type;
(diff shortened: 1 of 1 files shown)


Aingar:
The added check against ci.dwTargetShip is a performance booster, prevents further execution if we're colliding with clientside debris/asteroids.

Plugins/Public/playercntl_plugin/Main.cpp
```diff
@@ -588,9 +592,28 @@ namespace HkIServerImpl
 		returncode = DEFAULT_RETURNCODE;
 
 		// If spin protection is off, do nothing.
-		if (set_fSpinProtectMass == -1.0f)
+		if (!ci.dwTargetShip || set_fSpinProtectMass == -1.0f)
 			return;
```
Lazrius, line 596: Add braces

Lazrius (approved):

Plugins/Public/playercntl_plugin/Main.cpp
Aingar: done

dsyalex (approved):
Per Laz's approval, extra diff since looks good

PR #253, opened 2023-08-22
Pob building crafting rework

Aingar:
Moving #207 to a master branch PR

Copying previous, pretty outdated description. Will write a new one at  a future date.

Heavy rewrite of the entire facmod and buildmod. Heavy de-hardcodifying of both modules.

-Allows defining new buildings and recipes purely via Config (any code functionality for new non-factory buildings would still need to be developed in, obviously)
-The above allows us to dynamically define new Factory buildings with

Recipes are assigned to a 'crafting list', and factories can have access to any number of crafting lists as defined in their config.
For example, we can have a codename gun EXCALIBUR recipe added to 'snubcode1' crafting list, and then a Snub Codename Factory defined which has two crafting lists defined for it, 'snubcode1' and 'snubcode2', making the recipe available to craft if a factory is built.

Buildings are also separated into categories to better support their growing numbers. They are also segregated into 'building lists'. Currently there are three of them: 'core' containing core upgrade, shield generator and bulk storage, 'defense' which contains defense platforms and 'factories' containing everything else.

Other changes:

    'Smart' commands that automatically find the factory and recipe via name, one simply needs to type '/craft snubcode1 excalibur' to begin crafting of a recipe. Recipe numbers are still supported and are defined on per-craftlist basis
    To support Smart Commands, no more than 1 factory of a given type is allowed per POB.

I realize this PR is overly massive and packed with features, I might make an attempt to break it down once I manage to establish it fully works as intended.


Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -410,163 +410,137 @@ void CoreModule::SaveState(FILE *file)
...
+		{
+			if (!number_of_people)
+				break;
 
+			uint food_available = base->HasMarketItem(i.first);
+			if (food_available)
+			{
+				uint food_to_use = min(food_available, number_of_people);
+				base->RemoveMarketGood(i.first, food_to_use);
+				number_of_people -= food_to_use;
```
dsyalex, line 532: how about we name this `unfed_people`

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -410,163 +410,137 @@ void CoreModule::SaveState(FILE *file)
...
 			}
 		}
-		//else we do not change health, but we still need to send an update to fix the undockable problem. The base either has no logic or is invulnerable, so processing changes is useless.
-		else
+
+		// Insufficent food so fail check
```
dsyalex, line 536: what about killing off crew if there is insufficient food though?

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -28,18 +28,18 @@ PlayerBase::PlayerBase(uint client, const wstring &password, const wstring &the_
...
 	// Setup derived fields
 	SetupDefaults();
 
 	save_timer = rand() % 60;
 }
 
-PlayerBase::PlayerBase(const string &the_path)
+PlayerBase::PlayerBase(const string& the_path)
 	: path(the_path), base(0), money(0),
-	base_health(0), base_level(0), defense_mode(0), proxy_base(0), affiliation(0), siege_mode(false),
```
dsyalex, line 41: why is siege_mode(false) being removed here and what does it have to do with crafting?

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -314,7 +319,11 @@ void PlayerBase::Load()
 						else {
 							bp.admin = true;
 						}
-						passwords.push_back(bp);
+						passwords.emplace_back(bp);
+					}
+					else if (ini.is_value("crew_supplied"))
```
dsyalex, line 324: what if this never gets called e.g. because we're reading a version of the base file written by an older version of the plugin? won't that cause problems with the default of isCrewSupplied being false?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
+	}
+	else
+	{
+		openLine = L" - ";
+		closeLine = L"\n";
+	}
+	info += openLine + L"Pending " + stows(itos(build_queue.size())) + L" items";
+	if (active_recipe.nickname)
+	{
+		info += openLine + L"Crafting " + Status + active_recipe.infotext + L". Waiting for:";
```
dsyalex, line 45: missing closeLine at the end here?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
-			for (map<uint, uint>::iterator i = active_recipe.consumed_items.begin();
-				i != active_recipe.consumed_items.end(); ++i)
+			const GoodInfo* gi = GoodList::find_by_id(good);
+			if (gi)
+			{
+				info += openLine + L"- " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
+				if (quantity > 0 && base->HasMarketItem(good) < active_recipe.cooking_rate)
+				{
+					info += L" [Out of stock]";
+				}
```
dsyalex, line 59: missing `info += closeLine;`?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
+			}
+		}
+		if (active_recipe.credit_cost)
+		{
+			info += openLine + L" - Credits x" + stows(itos(active_recipe.credit_cost));
```
dsyalex, line 64: ditto comments about both closeLine and credits formatting

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
+			if (base->money < active_recipe.credit_cost)
+			{
+				info += L" [Insufficient cash]";
+			}
```
dsyalex, line 68: missing closeLine

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
+		}
+		vector<pair<uint, uint>> neededWorkforce;
+		if (!active_recipe.catalyst_items.empty())
+		{
+			info += openLine + L"Needed catalysts:";
```
dsyalex, line 73: missing closeLine

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
-		info += FACTORY_NAMES[type];
-		info += L" - Pending " + stows(itos(build_queue.size())) + L" items ";
-		if (active_recipe.nickname)
+		if (!neededWorkforce.empty())
 		{
-			info = L" - Building " + Status + active_recipe.infotext + L". Waiting for:";
-
-			for (map<uint, uint>::iterator i = active_recipe.consumed_items.begin();
-				i != active_recipe.consumed_items.end(); ++i)
+			info += openLine + L"Needed workforce:";
```
dsyalex, line 98: ditto

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
-					info += L"<PARA/><TEXT>      - " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
-					if (quantity > 0 && base->HasMarketItem(good) < active_recipe.cooking_rate)
-						info += L" [Out of stock]";
-					info += L"</TEXT>";
+					info += openLine + L" - " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
+					uint presentAmount = base->HasMarketItem(good);
+					if (presentAmount < quantity)
+					{
+						info += L" [Need " + stows(itos(quantity - presentAmount)) + L" more]";
+					}
```
dsyalex, line 92: ditto

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -32,53 +25,95 @@ wstring FactoryModule::GetInfo(bool xml)
...
+				const GoodInfo* gi = GoodList::find_by_id(good);
 				if (gi)
 				{
-					info += L" " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
+					info += openLine + L" - " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
+					uint presentAmount = base->HasMarketItem(good);
+					if (presentAmount < quantity)
+					{
+						info += L" [Need " + stows(itos(quantity - presentAmount)) + L" more]";
+					}
```
dsyalex, line 112: ditto

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -90,181 +125,196 @@ wstring FactoryModule::GetInfo(bool xml)
...
+		}
+		base->reservedCatalystMap[good] += quantityNeeded;
+	}
+
+	if (active_recipe.credit_cost)
 	{
-		uint good = i->first;
-		uint quantity = i->second > active_recipe.cooking_rate ? active_recipe.cooking_rate : i->second;
-		if (quantity)
+		uint moneyToRemove = min(active_recipe.cooking_rate * 10, active_recipe.credit_cost);
```
dsyalex, line 164: what is this min against `active_recipe.cooking_rate * 10` about?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -90,181 +125,196 @@ wstring FactoryModule::GetInfo(bool xml)
...
-					return false;
-				}
-			}
+		}
+	}
+
+	for (auto& i : active_recipe.consumed_items)
+	{
+		uint good = i.first;
+		uint quantity = min(active_recipe.cooking_rate, i.second);
```
dsyalex, line 179: is this min supposed to be here?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -90,181 +125,196 @@ wstring FactoryModule::GetInfo(bool xml)
...
+			continue;
+		}
+		cooked = false;
+		auto market_item = base->market_items.find(good);
+		if (market_item != base->market_items.end()
+			&& market_item->second.quantity >= quantity)
+		{
+			i.second -= quantity;
+			base->RemoveMarketGood(good, quantity);
+			return false;
```
dsyalex, line 191: why are we returning here? it sounds like we successfully took the commodity

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -90,181 +125,196 @@ wstring FactoryModule::GetInfo(bool xml)
...
+void FactoryModule::LoadState(INI_Reader& ini)
 {
 	active_recipe.nickname = 0;
+	RECIPE foundRecipe;
 	while (ini.read_value())
 	{
 		if (ini.is_value("type"))
 		{
-			type = ini.get_value_int(0);
+			factoryNickname = moduleNumberRecipeMap[ini.get_value_int(0)].nickname;
```
dsyalex, line 242: ditto one of my previous comments about indexing off input

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -282,5 +332,52 @@ bool FactoryModule::ToggleQueuePaused(bool NewState)
...
+	//return true if value changed
+	return RememberState != NewState;
+}
+
+FactoryModule* FactoryModule::FindModuleByProductInProduction(PlayerBase* pb, uint searchedProduct)
+{
+	for (auto& module : pb->modules)
+	{
+		FactoryModule* facModPtr = dynamic_cast<FactoryModule*>(module);
+		if (facModPtr && facModPtr->active_recipe.nickname == searchedProduct)
```
dsyalex, line 344: wouldn't it be safest to do a IsFactoryModule-like check on these first?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -282,5 +332,52 @@ bool FactoryModule::ToggleQueuePaused(bool NewState)
...
+	}
+	return nullptr;
+}
+
+void FactoryModule::StopAllProduction(PlayerBase* pb)
+{
+	for (auto& i : pb->modules)
+	{
+		FactoryModule* facModPtr = dynamic_cast<FactoryModule*>(i);
+		if (facModPtr)
```
dsyalex, line 357: ditto

dsyalex (commented):
have not looked at Main.cpp or PlayerCommands.cpp yet

Plugins/Public/base_plugin/BuildModule.cpp
```diff
@@ -171,48 +185,54 @@ bool BuildModule::Timer(uint time)
...
-			Paused = ini.get_value_bool(0);
-		}
-		else if (ini.is_value("produced_item"))
-		{
-			active_recipe.produced_item = ini.get_value_int(0);
-		}
-		else if (ini.is_value("cooking_rate"))
-		{
-			active_recipe.cooking_rate = ini.get_value_int(0);
+			active_recipe = moduleNumberRecipeMap[ini.get_value_int(0)];
```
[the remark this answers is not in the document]
Aingar: In case of a std::map or std::unordered_map, attempt to read a nonexistent element from a map simply calls a default constructor. I suppose I could still add some checks but what this code would result in would be simply loading an empty recipe that would instantly complete and disappear without adding anything to the base, effectively silently quitting.

Plugins/Public/base_plugin/BuildModule.cpp
```diff
@@ -171,48 +185,54 @@ bool BuildModule::Timer(uint time)
 	return false;
 }
 
-void BuildModule::LoadState(INI_Reader &ini)
+void BuildModule::LoadState(INI_Reader& ini)
```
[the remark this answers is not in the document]
Aingar: No, I specifically made sure not to use any new mandatory fields. Many have in fact been removed and their values are being populated at runtime from config.

Initial tests for reverse compatibility were successful but it's been a few month since I last ran them, there will be another batch of tests coming very soon when Haste has a bit of capacity.

Plugins/Public/base_plugin/ClientCommands.cpp
```diff
@@ -1,17 +1,17 @@
 #include "main.h"
 
 /// Send a command to the client at destination ID 0x9999
-void SendCommand(uint client, const wstring &message)
+void SendCommand(uint client, const wstring& message)
```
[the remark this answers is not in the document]
Aingar: No, but the code was near-unreadable as it was even for me, so it felt like a good opportunity to do some basic reformatting.

Plugins/Public/base_plugin/BuildModule.cpp
```diff
@@ -45,21 +37,40 @@ wstring BuildModule::GetInfo(bool xml)
 				info += L"</TEXT>";
 			}
 		}
+		if (active_recipe.credit_cost)
+		{
+			info += L"<PARA/><TEXT>      - Credits x" + stows(itos(active_recipe.credit_cost));
```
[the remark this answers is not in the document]
Aingar: will add in next commit

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -410,163 +410,137 @@ void CoreModule::SaveState(FILE *file)
...
 	}
 
-	if (space_obj)
+	if (!space_obj)
+	{
+		return false;
+	}
+
+	float rhealth = base->base_health / base->max_base_health;
+	pub::SpaceObj::SetRelativeHealth(space_obj, rhealth);
```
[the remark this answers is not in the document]
Aingar: It was very much done, it's the code extracted from a now non-existent else statement from line 562.
The purpose was to force the client to register the solar as a dockable and allow pressing F3 on it. The comment did not survive the 'transplant', will re-add

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -410,163 +410,137 @@ void CoreModule::SaveState(FILE *file)
...
+		base->base_health = base->max_base_health;
+	}
+	else if (base->base_health <= 0)
+	{
+		base->base_health = 0;
+	}
+
+	// Humans use commodity_oxygen, commodity_water. Consume these for
+	// the crew or kill 10 crew off and repeat this every 12 hours.
+	if (!dont_eat && time % set_crew_check_frequency == 0)
```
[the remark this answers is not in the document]
Aingar: It doesn't have much to do with it. It was a minor change that I didn't think twice about putting in.
However, I won't be cutting it out to put into a separate PR at this stage as we're too far gone for that.
Probably would be better to do with a bit of hindsight, but I simply didn't imagine the code would get the first review notes ~9 months after writing up the initial version.

Plugins/Public/base_plugin/CoreModule.cpp
Aingar: We're no longer doing that, instead opting for the crew to simply stop working if insufficiently supplied.
Feel free to poke Haste about this one.

Plugins/Public/base_plugin/PlayerBase.cpp
Aingar: Probably an honest mistake on my part, will re-add
Technically no functional change as bools initialize as false by default.

Plugins/Public/base_plugin/PlayerBase.cpp
Aingar: Since we have no way of knowing about 'Crew Status' prior to deployment of this patch, it'll have to be false until the first FOW check occurs after booting up the server post-patch.

It's a minor one-time inconvenience that we should be more than fine with as there are no real workarounds that I can see.

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: The idea is that openLine already contains a terminator for the last line (ever new line is preceded after a previous line that needs closing, after all. closeLine is only used at the very end of writing the string.

I have however noted that the newline prefix was not present in non-XML variant, adding.

Plugins/Public/base_plugin/CoreModule.cpp
```diff
@@ -410,163 +410,137 @@ void CoreModule::SaveState(FILE *file)
...
+	{
+		bool passedCheck = true;
```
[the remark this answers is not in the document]
Aingar: fixed in next commit

Plugins/Public/base_plugin/CoreModule.cpp
Aingar: peopleToFeed made more sense to me, added in next commit

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: If it's not a FactoryModule, the dynamc cast will return a nullptr, for which I already check.

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: It's in line with current behavior: If there is anything that had to be consumed, the recipe isn't done, and after consuming the cargo, we stop. If it was the very last chunk of commodities to eat, it'll have everything consumed and will be primed to finish crafting within the next tick (currently 20 seconds).

I could do a secondary check but having to iterate over the second time would be a non-trivial performance cost that I'd rather not opt in.

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: Yes, If for whatever reason the remaining commodities end up being lower than the cooking rate, we only want to consume the smaller of the two, so a recipe requiring 500 of Silvery_Metal and a cooking rate of 200 will consume 200+200+100 instead of 200+200+200

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: See above

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: See above and added credit formatting

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: We're consuming credits at 10x the cooking rate of everything else, otherwise it'd be painfully slow if we put in any reasonably large credit amount.

As for min, see above, we don't want to consume more than the recipe actually needs.

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -1,81 +1,119 @@
...
 	active_recipe.nickname = 0;
+	for (wstring& craftType : factoryNicknameToCraftTypeMap[factoryNickname])
+	{
+		base->availableCraftList.insert(craftType);
+	}
 }
 
 wstring FactoryModule::GetInfo(bool xml)
 {
 	wstring info;
```
dsyalex, line 22: doesn't this need to begin with `<TEXT>` if in XML mode?

Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -1,81 +1,119 @@
...
-					info += L" " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
+					info += openLine + L" - " + stows(itos(quantity)) + L"x " + HkGetWStringFromIDS(gi->iIDSName);
+					uint presentAmount = base->HasMarketItem(good);
+					if (presentAmount < quantity)
+					{
+						info += L" [Need " + stows(itos(quantity - presentAmount)) + L" more]";
+					}
 				}
 			}
 		}
```
dsyalex, line 119: doesn't the return below here need to add a final `</TEXT>` if in XML mode?

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -71,14 +61,30 @@ list<REPAIR_ITEM> set_base_repair_items;
...
+/// Maps of shortcut numbers to recipes to construct item.
+map<wstring, map<uint, RECIPE>> recipeCraftTypeNumberMap;
+map<wstring, map<wstring, RECIPE>> recipeCraftTypeNameMap;
+map<uint, vector<wstring>> factoryNicknameToCraftTypeMap;
+map<wstring, RECIPE> moduleNameRecipeMap;
+map<uint, RECIPE> moduleNumberRecipeMap;
+map<wstring, map<uint, RECIPE>> craftListNumberModuleMap;
+set<wstring> buildingCraftLists;
+
+void AddFactoryRecipeToMaps(const RECIPE& recipe);
```
dsyalex, line 86: we're just going to leave these functions here defined unimplemented near the top of a cpp file?

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2259,66 +2350,53 @@ void __stdcall HkCb_AddDmgEntry(DamageList *dmg, unsigned short sID, float& newH
...
-	if (args.find(L"testrecipe") == 0)
+	}
+	else if (args.find(L"testfacrecipe") == 0)
 	{
 		returncode = SKIPPLUGINS_NOFUNCTIONCALL;
 
 		RIGHT_CHECK(RIGHT_BASES)
 
-		uint client = HkGetClientIdFromCharname(cmd->GetAdminName());
+			uint client = HkGetClientIdFromCharname(cmd->GetAdminName());
```
dsyalex, line 2387: random indentation?

dsyalex (commented):
have not looked at PlayerCommands yet

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: All the necessary prefixes and suffixes are handled inside the SendBaseStatus method in ClientCommands.cpp

Plugins/Public/base_plugin/FactoryModule.cpp
Aingar: See above

Plugins/Public/base_plugin/Main.cpp
Aingar: Simply a c++ forward declaration so I don't need to clutter top of the file with the method and can stash it at the bottom.

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -20,103 +20,166 @@
...
-			L"<TEXT>Docking Rights: Whitelisted ships only.</TEXT><PARA/><PARA/>"
-			L"<TEXT>Defense Mode 2 - Logic: Blacklist > Whitelist > Faction Whitelist > IFF Standing.</TEXT><PARA/>"
-			L"<TEXT>Docking Rights: Anyone with good standing.</TEXT><PARA/><PARA/>"
-			L"<TEXT>Defense Mode 3 - Logic: Blacklist > Whitelist > Faction Whitelist > Hostile</TEXT><PARA/>"
-			L"<TEXT>Docking Rights: Whitelisted ships only.</TEXT><PARA/><PARA/>"
-			L"<TEXT>Defense Mode 4 - Logic: Blacklist > Whitelist > Faction Whitelist > Neutral</TEXT><PARA/>"
-			L"<TEXT>Docking Rights: Anyone with good standing.</TEXT><PARA/><PARA/>"
-			L"<TEXT>Defense Mode 5 - Logic: Blacklist > Whitelist > Faction Whitelist > Neutral</TEXT><PARA/>"
-			L"<TEXT>Docking Rights: Whitelisted ships only.</TEXT><PARA/><PARA/>"
+L"<TRA bold=\"true\"/><TEXT>/base rep [clear]</TEXT><TRA bold=\"false\"/><PARA/>"
```
dsyalex, line 82: duplicate of line 44?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -1074,179 +1141,180 @@ namespace PlayerCommands
...
-			if (recipes[module_nickname].reqlevel > base->base_level)
-			{
-				PrintUserCmdText(client, L"ERR Insufficient Core Level");
-				return;
-			}
+		if (!clients[client].admin)
+		{
+			PrintUserCmdText(client, L"ERR Access denied");
+			return;
+		}
```
dsyalex, line 1291: don't we have a function somewhere up above for these two checks?

Plugins/Public/base_plugin/PlayerCommands.cpp
```diff
@@ -2031,109 +2138,109 @@ namespace PlayerCommands
...
 					}
-					break;
+					else
+					{
+						ConPrint(L"Base too close to %ls, distance: %um", HkGetWStringFromIDS(idsName).c_str(), static_cast<uint>(distance));
+					}
+					return false;
 				}
+				break;
+			}
```
dsyalex, line 2233: what's going on here with the closing brackets/indentation?

dsyalex (commented):
Remaining PlayerCommands comments

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: First one is a case block closing, second is the entire switch block closing

Plugins/Public/base_plugin/PlayerCommands.cpp
Aingar: Yes, and I could have sworn I have applied that methoid where applicable a long, long time ago. Applying it once more.

dsyalex (approved):

PR #256, opened 2023-08-27
Hyperjump v3

Aingar:
-Rework Jump Drives to spawn temporary Jump Holes
-Jump Drive QoL command rework
-Jump Drive range fuel calculation rework
-Remade "POB JHs" into actual custom spawned Jump Holes
-Jump Hole randomization system. Currently used exclusively for Einhart system and it's surrounding Uncharted systems. Einhart will have 2 random jump holes leading back to Sirius, and 2 random jump holes leading to unique Uncharted systems. The server will randomize them depending on the config, but most likely it'll be every day or every second day.

Numerous changes facilitating the above such as:
-Removal of Survey module functionalities
-Rework of POB spawn mechanics
-Ability to despawn a POB without deleting/moving the file
-Ability to respawn a POB in runtime (will make it an admin command in the future)
-Sending a clienthook call to override the selected solar target system info
-Adjusting the POB data printing to .ini files, including changing the coordinate system to coordinate-or-object targeting, as well as printing systems as nickname hashes, while preserving the ability to read the numeric hash for compatibility reasons

Miscellaneous changes such as:
-Rework of HkGetClientIDByShip to use COBject::Find method instead of iterating over player data array.
-Adding PvP damage protection when switching systems
-Minor refactor of how pvp invulnerability works
-Updating HkUnLightFuse method signature to describe the previously unknown 3rd parameter


Plugins/Public/base_plugin/CreateSolar.cpp
```diff
@@ -1,17 +1,34 @@
...
+void CreateSolar::DespawnSolarCallout(DESPAWN_SOLAR_STRUCT* info)
+{
+	if (!customSolarList.count(info->spaceObjId))
+	{
+		return;
+	}
+
+	if (pub::SpaceObj::ExistsAndAlive(info->spaceObjId) == 0)
+	{
+		pub::SpaceObj::Destroy(info->spaceObjId, info->destroyType);
```
dsyalex, line 12: We destroy if it either doesn't exist or is not alive?
Wonder if the return codes of ExistsAndAlive should be better shown here, or if this is just wrong

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -195,7 +191,16 @@ void PlayerBase::Load()
 					}
 					else if (ini.is_value("system"))
 					{
-						system = ini.get_value_int(0);
+						string sysNickname = ini.get_value_string(0);
```
dsyalex, line 194: ??? so this field now has two valid types, meaning either a string nickname or its hash? why would we need to store system as anything other than a hash?

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -211,9 +216,32 @@ void PlayerBase::Load()
...
 					}
+					else if (ini.is_value("destobject"))
+					{
+						destObjectName = ini.get_value_string(0);
+						destObject = CreateID(destObjectName.c_str());
+					}
 					else if (ini.is_value("destsystem"))
 					{
-						destsystem = ini.get_value_int(0);
+						string sysNickname = ini.get_value_string(0);
```
dsyalex, line 226: ditto

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -394,14 +416,30 @@ void PlayerBase::Save()
...
 		fprintf(file, "pos = %0.0f, %0.0f, %0.0f\n", position.x, position.y, position.z);
 
-		fprintf(file, "destsystem = %u\n", destsystem);
-		fprintf(file, "destposition = %0.0f, %0.0f, %0.0f\n", destposition.x, destposition.y, destposition.z);
-
 		Vector vRot = MatrixToEuler(rotation);
 		fprintf(file, "rot = %0.0f, %0.0f, %0.0f\n", vRot.x, vRot.y, vRot.z);
+		if (mapArchs[basetype].ishubreturn)
+		{
+			const auto& systemInfo = Universe::get_system(destSystem);
```
dsyalex, line 427: we should probably call this destSysInfo

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -394,14 +416,30 @@ void PlayerBase::Save()
...
+			fprintf(file, "destsystem = %s\n", systemInfo->nickname);
+
+			fprintf(file, "destpos = %0.0f, %0.0f, %0.0f\n", destPos.x, destPos.y, destPos.z);
+
+			Vector destRot = MatrixToEuler(destOri);
+			fprintf(file, "destori = %0.0f, %0.0f, %0.0f\n", destRot.x, destRot.y, destRot.z);
+		}
+		else if (mapArchs[basetype].isjump && destObject && pub::SpaceObj::ExistsAndAlive(destObject) == 0)
```
dsyalex, line 435: ExistsAndAlive return code weirdness again - assume this is an enum somewhere?

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -394,14 +416,30 @@ void PlayerBase::Save()
...
+		{
+			uint systemId;
```
dsyalex, line 437: destSystemId

Plugins/Public/base_plugin/PlayerBase.cpp
```diff
@@ -394,14 +416,30 @@ void PlayerBase::Save()
...
+			pub::SpaceObj::GetSystem(destObject, systemId);
+			const auto& systemInfo = Universe::get_system(systemId);
```
dsyalex, line 439: destSystemInfo

dsyalex (changes requested):
This desperately needs a PR description. It is a monster of a PR with 120 or so commits, touching everything from FLHook Core & the SDK, PluginUtilities, the POB plugin, the mining control plugin, the player control plugin, etc. and all we have for it as a summary is the title 'Hyperjump v3'. I've also spotted various changes included in here which do not look necessary to this PR and which therefore should almost certainly be called out in the description.

Have not looked at the following yet:
* Plugins/Public/playercntl_plugin/Main.cpp
* Plugins/Public/base_plugin/Main.cpp
* Plugins/Public/base_plugin/Jumper.cpp
* Plugins/Public/base_plugin/HyperJump.cpp

Plugins/Public/base_plugin/CreateSolar.cpp
Aingar: 0 means alive, -2 means dead, no other values are returned. I've added this exact note to several other places this method is called, but I guess I missed this one.

Plugins/Public/base_plugin/PlayerBase.cpp
Aingar: Again, it returns 0 for alive, -2 for dead. I tried to make it return an enum internally but that wasn't an option. In hindsight I should just make a wrapper function interpreting the value and returning appropriate boolean.

Source/FLHook/HkFuncTools.cpp
```diff
@@ -216,12 +216,11 @@ HK_ERROR HkResolveShortCut(const wstring &wscShortcut, uint &_iClientID)
 
 uint HkGetClientIDByShip(uint iShip)
 {
-	for (uint i = 0; i <= MAX_CLIENT_ID; i++)
+	const CShip* cobj = reinterpret_cast<CShip*>(CObject::Find(iShip, CObject::CSHIP_OBJECT));
```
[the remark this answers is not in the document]
Aingar: For a single player ingame, results in miliseconds for a million repeats

Search for a nonexistent/ non-player ship:
timeCOBjectFind: 9
standardSearch: 150

Search for an existing player ship. Do note that the old search scales linearly with half of amount of players logged in, so it would take ~60 players online to catch up.
timeCOBjectFind: 62
standardSearch: 2

Plugins/Public/base_plugin/PlayerBase.cpp
Aingar: For convenience of admins reading/writing to those files. If human-readability wasn't a factor, we'd use single character variable names to save on space and performance :]

Plugins/Public/playercntl_plugin/Main.cpp
```diff
@@ -428,6 +437,18 @@ namespace HkIServerImpl
...
+	void __stdcall RequestCancel(int iType, unsigned int iShip, unsigned int p3, unsigned long p4, unsigned int iClientID)
+	{
+		returncode = DEFAULT_RETURNCODE;
+
+		if (!iClientID)
+		{
+			return;
+		}
+
+		HyperJump::RequestCancel(iType, iShip, p3, p4);
```
dsyalex, line 449: we not also going to send iClientID here?

also, p3 has a name in the called function, so I expect it to have a name here too

Plugins/Public/playercntl_plugin/Main.cpp
```diff
@@ -620,10 +646,12 @@ namespace HkIServerImpl
 
 	void __stdcall SystemSwitchOut(uint iClientID, FLPACKET_SYSTEM_SWITCH_OUT& switchOutPacket)
 	{
-		// in case of SERVER_PACKET hooks, first argument is junk data before it gets processed by the server.
-		uint packetClient = HkGetClientIDByShip(switchOutPacket.shipId);
-		if(packetClient)
-			AntiJumpDisconnect::SystemSwitchOut(packetClient);
+		if (iClientID != HkGetClientIDByShip(switchOutPacket.shipId))
```
dsyalex, line 649: Backwards, I think this comparison is

Plugins/Public/playercntl_plugin/Main.cpp
```diff
@@ -1207,29 +1235,12 @@ USERCMD UserCmds[] =
...
-	{ L"/jump*",		HyperJump::UserCmd_ActivateJumpDrive, L"Usage: /jump"},
-	{ L"/beacon",		HyperJump::UserCmd_DeployBeacon, L"Usage: /beacon"},
-	{ L"/beacon*",		HyperJump::UserCmd_DeployBeacon, L"Usage: /beacon"},
-	{ L"/jumpbeacon",		HyperJump::UserCmd_JumpBeacon, L"Usage: /jumpbeacon"},
-	{ L"/jumpbeacon*",		HyperJump::UserCmd_JumpBeacon, L"Usage: /jumpbeacon"},
-	{ L"/charge",		HyperJump::UserCmd_ChargeJumpDrive, L"Usage: /charge"},
-	{ L"/charge*",		HyperJump::UserCmd_ChargeJumpDrive, L"Usage: /charge"},
-	{ L"/jumpsys",		HyperJump::UserCmd_ListJumpableSystems, L"Usage: /jumpsys"},
+	{ L"/setsector",	HyperJump::UserCmd_SetSector, L"Usage: /setsector <number>"},
+	{ L"/jump",			HyperJump::UserCmd_Jump, L"Usage: /jump <systemName/blind/stop/list>"},
```
dsyalex, line 1239: indentation didn't work out here

Plugins/Public/playercntl_plugin/Main.cpp
```diff
@@ -1207,29 +1235,12 @@ USERCMD UserCmds[] =
...
+	{ L"/jumpbeacon",	HyperJump::UserCmd_JumpBeacon, L"Usage: /jumpbeacon <playername/playerID>" },
```
dsyalex, line 1240: IIRC it's customary to allow:
`/jumpbeacon <playername>`
to also be called with
`/jumpbeacon$ <playerID>`
but I don't think it usually went in the descriptions?

dsyalex (commented):
playercntl main

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -922,6 +936,21 @@ void LoadSettingsActual()
...
 	}
 
+	// loadHyperspaceHubConfig is weekday where 0 = sunday, 6 = saturday
+	// if it's today, randomize appropriate 'jump hole' POBs
+	if (bmapLoadHyperspaceHubConfig)
+	{
+		time_t tNow = time(0);
+		struct tm *t = localtime(&tNow);
+		uint currWeekday = (t->tm_wday + 6)%7; // conversion from sunday-week-start to monday-start
+		if (bmapLoadHyperspaceHubConfig & (1 << currWeekday)) // 1 - monday, 2 - tuesday, 4 - wednesday and so on
```
dsyalex, line 946: There has *got* to be a better way to let admins configure this than bitwise logic against a number that's going to look very magic to them. I think we should turn the `bmapLoadHyperspaceHubConfig = ini.get_value_int(0);` above into:
```c++
bmapLoadHyperspaceHubConfig = 0;
vector<str> weekdays;
weekdays.emplace_back("monday");
weekdays.emplace_back("tuesday");
weekdays.emplace_back("wednesday");
weekdays.emplace_back("thursday");
weekdays.emplace_back("friday");
weekdays.emplace_back("saturday");
weekdays.emplace_back("sunday");
for (int i = 0; i < 7; i++) {
    str val = ToLower(ini.get_value_string(i));
    vector<str>::iterator loc = std::find(weekdays.begin(), weekdays.end(), val);
    if (loc != weekdays.end()) {
        bmapLoadHyperspaceHubConfig |= (1 << (loc + 1));
    }
}
```
(disclaimer: wrote this off the top of my head, no idea if it so much as compiles)

Plugins/Public/base_plugin/Main.cpp
```diff
@@ -2661,50 +2667,45 @@ bool ExecuteCommandString_Callback(CCmds* cmd, const wstring &args)
...
 		}
 		wstring type = cmd->ArgStr(3);
 		if (!type.length())
 		{
 			PrintUserCmdText(client, L"ERR No type");
-			PrintUserCmdText(client, L"Usage: .jumpcreate <archtype> <loadout> <type> <dest system> <x> <y> <z> <affiliation> <name>");
+			PrintUserCmdText(client, usage.c_str());
 			return true;
 		}
 		wstring destsystem = cmd->ArgStr(4);
```
dsyalex, line 2687: should we not be renaming this?

dsyalex (commented):
base main

Plugins/Public/playercntl_plugin/Main.cpp
Aingar: Problem with RequestEvent and RequestCancel is that the parameters have different meanings depending on the value of the first parameter, hence why I haven't named the variables in the main.cpp, but did in the HyperJump subcall, where we only care about a very specific event type == 0 aka dock request

Plugins/Public/playercntl_plugin/Main.cpp
Aingar: I am aware, but given how we don't allow players to be named '231' or '66', I don't see why wouldn't we just have it under a single command.

Plugins/Public/playercntl_plugin/Main.cpp
Aingar: Well, I have to say that it looks as it should as far as I can tell, GitHub being funny? obraz

Plugins/Public/playercntl_plugin/Main.cpp
Aingar: I think it's neater when the variable and name of the function are next to each, as opposed to
`if (HkGetClientIDByShip(switchOutPacket.shipId) != iClientID)`
where iClientID at a glance is compared with switchOutPacker.shipID

Plugins/Public/base_plugin/Main.cpp
Aingar: Implemented a bitmap populating pattern used in other plugins.

Plugins/PluginUtilities/PluginUtilities.cpp
```diff
@@ -1281,7 +1281,7 @@ __declspec(naked) void __stdcall HkLightFuse(IObjRW *ship, uint iFuseID, float f
 	}
 }
 
-__declspec(naked) void __stdcall HkUnLightFuse(IObjRW *ship, uint iFuseID, float fDunno)
+__declspec(naked) void __stdcall HkUnLightFuse(IObjRW *ship, uint iFuseID, float fDelay)
```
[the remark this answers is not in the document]
Aingar: added to the top comment

dsyalex (commented):
no comments for Jumper.cpp

Plugins/Public/playercntl_plugin/HyperJump.cpp
```diff
@@ -475,262 +525,575 @@ namespace HyperJump
...
 
-			if (!bc.decayed)
+		auto& systemRangeList = mapAvailableJumpSystems[iSystemID];
+		for (uint depth = 1; depth <= playerJumpDrive.arch->jump_range; depth++)
+		{
+			wstring sysList = L"Systems " + stows(itos(depth)) + L" jumps away:";
+			for (uint &allowed_sys : systemRangeList[depth])
+			{
+				const Universe::ISystem *systemData = Universe::get_system(allowed_sys);
+				sysList += L" [" + HkGetWStringFromIDS(systemData->strid_name) + L"]";
```
dsyalex, line 896: this looks like a funny way of building a list to show to humans

Plugins/Public/playercntl_plugin/HyperJump.cpp
```diff
@@ -744,225 +1107,205 @@ namespace HyperJump
...
+								//bump up timer to execute the jump as soon as beacon player exits the jump tunnel.
+								jd.jump_timer++;
+								continue;
 							}
+
+							uint targetShip;
+							pub::Player::GetShip(jd.targetClient, targetShip);
+							pub::SpaceObj::GetLocation(targetShip, jd.vTargetPosition, jd.matTargetOrient);
+							pub::Player::GetSystem(jd.targetClient, jd.iTargetSystem);
+							int innacurracy = mapPlayerBeaconMatrix[jd.targetClient].arch->inaccuracy;
```
dsyalex, line 1186: inaccuracy

dsyalex (approved):

PR #260, opened 2023-09-05
Cloak: Prevent disabling of cloak during activation sequence. Fix sub-zero fuel consumption logic.

Aingar:
Done to prevent being able to activate cloak, then deactivate it to achieve a permanent transparency on your ship. The deactivation fails if done by a command, and in case of running out of fuel, is delayed until the sequence finishes.
Also, minor code optimization.


dsyalex (changes requested):
see inline comment

dsyalex (approved):

PR #261, opened 2023-09-05
KillTracker: Remove unwanted on-disconnect logic, fix solution/project files

Aingar:
Originally this PR adjusted the disconnect hook functionality, but now instead trims it down to the bare minimum after a chat with Haste. Also noticed that project files must have gotten messed up somewhere along the way as they were missing/malformed in the solution file.


dsyalex (changes requested):
Rationale for modification of Disconnect function is missing

PR #262, opened 2023-09-08
Block solars from buying specific commodities

Aingar:
Blocks selected stations from buying selected goods.
Used to prevent stations selling refined ore products at a premium from being a valid sellpoint from POB-acquired commodities.
For example: Selling gold to Arranmore.

### DIFF
--- a/Plugins/Public/commoditylimit/Main.cpp
+++ b/Plugins/Public/commoditylimit/Main.cpp
@@ -68,13 +68,6 @@ struct CommodityLimitStruct
 	//list<uint> ShipClassRestrictions;
 };
 
-struct SellReverse
-{
-	uint archId;
-	float status;
-
-};
-
 unordered_map<uint, CommodityLimitStruct> mapCommodityRestrictions;
 map<uint, bool> mapBuySuppression;
 unordered_map<uint, unordered_set<uint>> mapProducers;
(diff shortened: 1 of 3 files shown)


Plugins/Public/commoditylimit/Main.cpp
```diff
@@ -145,6 +166,45 @@ void ClearClientInfo(uint iClientID)
...
+{
+	returncode = DEFAULT_RETURNCODE;
+	if (!sellSuppressionMap.count(client))
+	{
+		return;
+	}
+
+	uint baseId = Players[client].iBaseID;
+	uint goodId = sellSuppressionMap.at(client);
+	float price = BaseDataList_get()->get_base_data(baseId)->market_map.find(goodId).value()->fPrice;
```
dsyalex, line 198: How well does this interact with prices.cfg overrides from DSAceSrv?

Also, don't we have pre-existing buying blocking code somewhere?

Aingar: DSAceSrv updates the BaseDataList, so there is no issue here.

As for the code existing elsewhere, I found it in Alley's plugin. Moving code there.

Aingar: Note to self: Don't code past 3AM.

Alley's code was for buy prevention, not sell prevention. This code is Sell prevention, which is only present in POB plugin, and it'd be a very poor place to force this feature into.

dsyalex (approved):

PR #277, opened 2023-09-29
POBs: Indestructible POBs don't process damage taken.

Aingar:
Results in no wasted CPU cycles, and no shields being triggered. win-win.


dsyalex (approved):

dsyalex:
Maybe we'll want jump holes raising their shields in future, but not now

PR #278, opened 2023-10-01
BuyBackBlock: Simply block all bases selling a defined good from buying it.

Aingar:
Value of stock == 500 is used globally for every base selling a given commodity.


dsyalex:
why

dsyalex (approved):

PR #284, opened 2023-10-03
(MustHave)POBCraft: Fix issues with refineries

Aingar:
-incorrect handling of factory building
-unintuitive configs for refinery modules
-craft lists not being cleared on module destruction -craft lists not being selectable via numbers
-various optimizations, including food consumption and FactoryModule saving.


Plugins/Public/base_plugin/FactoryModule.cpp
```diff
@@ -258,7 +259,7 @@ void FactoryModule::LoadState(INI_Reader& ini)
 	{
 		if (ini.is_value("type"))
 		{
-			factoryNickname = moduleNumberRecipeMap[ini.get_value_int(0)].nickname;
+			factoryNickname = CreateID(ini.get_value_string(0));\
```
dsyalex, line 262: random backslash? also, why is factoryNickname seemingly a uint?

dsyalex (changes requested):

dsyalex (approved):

PR #291, opened 2023-10-21
Preset message macro updates

danswann:
# Description
1) Adds a new macro `#r` available in `/setmsg` that expands to the remaining text when invoking the preset, for example:
```
/setmsg 0 We're on patrol from #r.
/0 Planet Manhattan
Player: We're on patrol from Planet Manhattan.
```

2) Updates the existing `#t` macro to support targeting dockable bases in addition to player character names. Example:
```
/setmsg 0 We're headed to #t.
*player clicks on Norfolk Shipyard*
/0
Player: We're headed to Norfolk Shipyard.
```

# Design Decisions
### All or Nothing
For `#r` I thought about allowing multiple arguments, like
```
/setmsg 0 We're hauling #1 from #2.
/0 Cardamine Malta
Player: We're hauling Cardamine from Malta.
```
but -- like most command parsing -- that quickly becomes an issue when your arguments have spaces in them.  
Obviously it's possible (quotes, a different delimiter, etc.) but I chose to favor simplicity.

# Limitations
### Target Viability
The update to `#t` only works for space objects that have associated dockables (bases, dockable planets, docking rings, etc.).
I couldn't find a way to get the display name of other space objects (like uninhabited planets), the closest I could get was the solar archetype name ("Planet").  
If someone knows how to do that, do let me know. Documentation for FLHook seems... sparse. I'm mostly just looking at method signatures and existing code and making guesses.
### Testing
Since Patch 5.0 Hotfix 1 I haven't been able to connect to a server running on the same machine as the client, so my *actual* multiplayer testing (i.e. more than 1 character online at once) was limited to multiboxing on my laptop briefly to make sure `#t` still worked as intended on other players. I imagine that's probably sufficient, as these change seems pretty straightforward, but maybe someone wants to test it with more rigor.

# Other Considerations
The `#r` macro does allow people to send messages that are nearly twice as long as the default chatbox limit (i.e. a maximally long preset combined with some maximally long replacement text).  
This didn't seem to cause any issues in my brief testing, but I don't know if there's other stuff Discovery uses that makes assumptions about chat message length that might be affected. Just something to think about.


danswann:
> Since Patch 5.0 Hotfix 1 I haven't been able to connect to a server running on the same machine as the client

I got this fixed with the anti-IPV6 patch and have been able to do more testing. I found a bug with `/t[0-9]` because it's implemented differently than all the rest, will commit a fix in a bit.

I also noticed that `/r[0-9]` are just not registered at all, so presets cannot be used in replies. Is that intentional?

danswann:
Fixed the issue with `/t[0-9]` but accidentally merged without a rebase.

danswann:
Reset to before erroneous merge and rebased on latest upstream changes.


So for most message types (system, local, group) the regular message commands and the preset message commands are handled by different functions, for example `/g` and `/group` go to `Message::UserCmd_GroupMsg`, but `/gX` goes to `Message::UserCmd_GMsg`.

This is not the case for sending messages to a target though, as `/t`, `/target`, and `/tX` are *all* handled by `Message::UserCmd_SendToLastTarget`.

I decided to just do a quick fix to get `/tX` working correctly. While it would be nice to have all the different message types handled consistently, I'd rather do that in a separate refactor. `Message.cpp` could probably have it's LOC cut in half.

Aingar:
Hey, just letting you know it's not forgotten. Right now the codebase is frozen for the purpose of patching up any crashes. We'll get back to you :]

danswann:
Take all the time you guys need.

Worked out for the best anyway, cuz I just found and fixed a sneaky bug of my own.

Plugins/Public/playercntl_plugin/Message.cpp
```diff
@@ -356,6 +385,17 @@ namespace Message
 			wscMsg = ReplaceStr(wscMsg, L"#c", wscCurrLocation.c_str());
 		}
 
+		if (wscMsg.find(L"#r") != -1)
+		{
+			if (wscParam.length() == 0)
```
Aingar, line 390:
```suggestion
			if (wscParam.empty())
```

Plugins/Public/playercntl_plugin/Message.cpp
```diff
@@ -968,14 +1009,19 @@ namespace Message
...
 		}
-		else if (wscMsg.size() == 0)
+		// If this is a /t or /target command then send message text directly
+		else
 		{
-			PrintUserCmdText(iClientID, L"ERR Invalid parameters");
-			PrintUserCmdText(iClientID, usage);
-			return true;
+			wscMsg = GetParamToEnd(wscParam, ' ', 0);
+			if (wscMsg.size() == 0)
```
Aingar, line 1019:
```suggestion
			if (wscMsg.empty())
```

Aingar (changes requested):
Two minor requests. Also from what I understand the code, all the the #r tags would be replaced by the same value, as I don't see any for loops that would iterate over all provided parameters? Please elaborate on that.

danswann:
> Also from what I understand the code, all the the #r tags would be replaced by the same value, as I don't see any for loops that would iterate over all provided parameters? Please elaborate on that.

Sure. I talked about it in the "Design Decisions" section of the first PR comment, but the tl;dr is that I thought supporting multiple `#r`s would make it too cumbersome to use when you had parameters that themselves contained spaces.

As currently implemented, `#r` is just equivalent to everything-after-the-`/l[0-9]`. This is less powerful than supporting multiple `#r`s, but easier to use, and I think the general goal here is "help people type more quickly."

### Example
```
/setmsg 0 It's illegal to carry #r while inside #r.
/0 "Black Market Munitions" "Liberty house space"
Player: It's illegal to carry Black Market Munitions while inside Liberty house space.
```
This is powerful, but can become cumbersome. (Now what do you do when your parameters contain `"`?)

```
/setmsg Captain James: #r.
/0 We'll see what all this "pirate activity" is really about.
Player: Captain James: We'll see what all this "pirate activity" is really about.
```
This is less powerful, but dead simple to use (don't have to worry about using a space or using your delimiter character).


If you disagree and think multiple-`#r` support is worth that complexity, or if you can think of a more user-friendly way to implement it than I could, I'm happy to modify the code to support it.

Plugins/Public/playercntl_plugin/Message.cpp
danswann: Change implemented.

Plugins/Public/playercntl_plugin/Message.cpp
danswann: Change implemented.

Aingar (approved):
Can always look into making #r more robust in the future. Happy with what I see for now.

PR #292, opened 2023-10-21
Enhanced performance logging

Aingar:
Allows for real time sampling of performance timers on both global scale and on individual hook scale

### DIFF
--- a/Source/FLHook/Hook.h
+++ b/Source/FLHook/Hook.h
@@ -62,7 +62,6 @@
 #define ADDR_COMMON_VFTABLE_MUNITION 0x139CE8
 #define ADDR_COMMON_VFTABLE_ENGINE 0x139AAC
 
-
 #define HK_GET_CLIENTID(a, b) \
 	bool bIdString = false; \
 	if(b.find(L"id ") == 0) bIdString = true; \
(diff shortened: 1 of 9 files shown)


Source/FLHook/Settings.cpp
```diff
@@ -31,6 +31,10 @@ bool			set_bLogUserCmds;
 bool			set_bPerfTimer;
 uint			set_iTimerThreshold;
 uint			set_iTimerDebugThreshold;
+bool			set_logPerfTimers;
+int				set_perfTimerLength;
+int				set_hookPerfTimerLength;
```
dsyalex, line 36:
```suggestion
int			set_perfTimerLength;
int			set_hookPerfTimerLength;
```

Source/FLHook/Hook.h
```diff
@@ -160,9 +160,13 @@ struct PLUGIN_SORTCRIT {
 			if(itplugin->pFunc) { \
 				CTimer timer(itplugin->sPluginFunction,set_iTimerThreshold); \
 				timer.start(); \
+				auto timeHookStart = std::chrono::high_resolution_clock::now();\
 				try { \
 					vPluginRet = ((ret_type (calling_convention*) arg_types )itplugin->pFunc) args; \
 				} catch(...) { AddLog("ERROR: Exception in plugin '%s' in %s", itplugin->sName.c_str(), __FUNCTION__); LOG_EXCEPTION } \
+				if(set_hookPerfTimerLength && set_perfTimedHookName == __FUNCTION__) \
```
dsyalex, line 167: do we not have a quicker way to check this than a string comparison?

dsyalex (approved):
minor stuff inline

Source/FLHook/Hook.h
Aingar: Afraid not, but note that it's prefaced with an UINT comparison which if failed (and it will fail unless logging is enabled), prevents the string comparison to begin with.

dsyalex (approved):

PR #296, opened 2023-10-21
Enable compiler optimizations

Aingar:


dsyalex (changes requested):
As discussed, PlayerCommands changes need to be moved to a separate PR

dsyalex (approved):

PR #304, opened 2023-10-31
Port better logging / minidump generation from Starport SVN

JohnWildkins:
title

not tested because it still doesn't compile lol


Aingar (commented):
Aingar's Stamp Of Approval

Aingar (approved):

PR #316, opened 2023-11-02
PVEController bounty drops can now randomize in a range.

Aingar:


dsyalex (changes requested):
drop.uAmountDroppedMax is already set, finalAmount -> amountDropped is just a rename, and this has a new bug in that amountDropped is not passed to the Server.MineAsteroid call

PR #326, opened 2023-11-04
Hook timer logging now toggleable via #define in hook.h

Aingar:
Also removed old timer methods that aren't useful.

### DIFF
--- a/Source/FLHook/Hook.h
+++ b/Source/FLHook/Hook.h
@@ -13,6 +13,8 @@
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////
 // defines
 
+//#define HOOK_TIMER_LOGGING
+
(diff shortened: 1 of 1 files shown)


Source/FLHook/Hook.h
```diff
@@ -285,6 +279,98 @@ struct PLUGIN_SORTCRIT {
...
+	if(bPluginReturn) \
+		return vPluginRet; \
+} \
+
+// same for void types, not really seeing a way to integrate it in 1st macro :(
+#define CALL_PLUGINS_V(callback_id,calling_convention,arg_types,args) \
+{ \
+	bool bPluginReturn = false; \
+	g_bPlugin_nofunctioncall = false; \
+	auto timeStart = std::chrono::high_resolution_clock::now();\
```
dsyalex, line 320: needs removal - and in the other two

dsyalex (approved):

PR #336, opened 2023-11-07
Logging: logged player death systems are human readable.

Aingar:
Admin Logging player deaths now saves system the death occured in as nickname instead of a numeric hash.

Also removed the incorrectly working BaseLand hook that is replaced by Dock_Call hook that is finalized in #339


dsyalex (changes requested):
This PR does more than it says on the tin. Also, since we're apparently abolishing CUSTOM_POB_DOCK_ALERT, shouldn't we get rid of it from the other places it's used too?

PR #339, opened 2023-11-07
Fix dock message bug for good.

Aingar:
PlayerCntl: move the message print to a separate check where dockPort == -1
MobileDock: move the returncode override to after it's confirmed the target is a player.


dsyalex (changes requested):
requires rebase

dsyalex (approved):

PR #341, opened 2023-11-07
Exception fix in /base Defmod command

Aingar:
Prevent dereference of a nullptr aka empty module in /base defmod method. Replaced with a dynamic_cast.


dsyalex (approved):
problem was:

> empty module slots
> it tried to dereference them and grab their type value
> when they were nullptrs

PR #351, opened 2023-11-25
Conn update, conn-POB removal, fixes.

Aingar:
-Revert to updating the player file upon inputting /conn, as doing so after they launch can randomly fail. 
-Can no longer /conn from POBs to prevent ammo-related abuse 
-Also minor code cleanup.

### DIFF
--- a/Plugins/Public/conn_plugin/Main.cpp
+++ b/Plugins/Public/conn_plugin/Main.cpp
@@ -174,23 +174,12 @@ bool ValidateCargo(unsigned int client)
 
 void StoreCurrentBase(uint client)
 {
-	CUSTOM_BASE_IS_DOCKED_STRUCT info;
-	info.iClientID = client;
-	info.iDockedBaseID = 0;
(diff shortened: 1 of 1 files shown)


Plugins/Public/conn_plugin/Main.cpp
```diff
@@ -254,11 +227,7 @@ void MoveClient(unsigned int client, unsigned int targetBase)
 			return;
 		}
 		PrintUserCmdText(client, L"Player base renamed/destroyed, ship redirected to a proxy base");
-		pub::Player::ForceLand(client, proxyBaseID); // beam
-		if (returnSys != system)
-		{
-			SimulateF1(client, proxyBaseID);
-		}
+		HkBeamById(client, targetBase);
```
dsyalex, line 230: We know this can't work because in this case Universe::get_base(targetBase) failed

```suggestion
		HkBeamById(client, proxyBaseID);
```

dsyalex (changes requested):

PR #357, opened 2023-11-25
Minor fixes

Aingar:
MobileDock: Fix /jettisonship handling
MunitionCntl: Fix fetched CGuided objects not being released
BountyScan: Add /bs shorthand command

### DIFF
--- a/Plugins/Public/munitioncntl/Main.cpp
+++ b/Plugins/Public/munitioncntl/Main.cpp
@@ -171,7 +171,8 @@ void ProcessGuided(FLPACKET_CREATEGUIDED& createGuidedPacket)
 	{
 		case NOTRACK_NOALERT:
 		{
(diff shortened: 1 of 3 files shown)


Plugins/Public/munitioncntl/Main.cpp
```diff
@@ -171,7 +171,8 @@ void ProcessGuided(FLPACKET_CREATEGUIDED& createGuidedPacket)
 	{
 		case NOTRACK_NOALERT:
 		{
-			const auto& projectile = reinterpret_cast<CGuided*>(CObject::Find(createGuidedPacket.iProjectileId, CObject::CGUIDED_OBJECT));
+			CGuided* projectile = reinterpret_cast<CGuided*>(CObject::Find(createGuidedPacket.iProjectileId, CObject::CGUIDED_OBJECT));
+			projectile->Release();
```
dsyalex, line 175: Shouldn't this be after we call set_target to ensure we're not about to try to interact with free'd memory?

dsyalex (changes requested):
.

